PR #17, opened 2017-10-05
r/heroku_app: always read all config vars

catsby:
The current implementation of `resource_heroku_app` only reads values for configuration variables for values defined in the current Terraform configuration files. This is counter to how Terraform Resources generally work. Unless there is an extreme circumstance, Terraform Resources should read all supported values regardless if they are set locally, so that a user can detect configuration drift. 

Consider this config:

```hcl
provider "heroku" {}

resource "heroku_app" "test_config_vars" {
  name   = "test-config-vars"
  region = "us"

  config_vars {
    OTHER_TOKEN = "twothree4"
    LAST_TOKEN  = "threefour5"
    EDIT        = "ok"
    NEW         = "best"
    MORE        = "another"
  }
}
```

If another user were to add additional configuration variables via the Web or API outside of Terraform, the current implementation would not detect it. I found this by importing an app with vars and none in my config, and no diff was detected.

In this patch, we read and save all configuration values. The `all_config_vars` (?) attribute should be removed in a future release. 

Test results after the patch:

```
TF_ACC=1 go test ./heroku -v -run=TestAccHerokuApp -timeout 120m
=== RUN   TestAccHerokuApp_importBasic
--- PASS: TestAccHerokuApp_importBasic (2.68s)
=== RUN   TestAccHerokuApp_importOrganization
--- PASS: TestAccHerokuApp_importOrganization (2.62s)
=== RUN   TestAccHerokuAppFeature
--- PASS: TestAccHerokuAppFeature (4.52s)
=== RUN   TestAccHerokuApp_Basic
--- PASS: TestAccHerokuApp_Basic (2.33s)
=== RUN   TestAccHerokuApp_NameChange
--- PASS: TestAccHerokuApp_NameChange (4.12s)
=== RUN   TestAccHerokuApp_NukeVars
--- PASS: TestAccHerokuApp_NukeVars (3.61s)
=== RUN   TestAccHerokuApp_Buildpacks
--- PASS: TestAccHerokuApp_Buildpacks (5.53s)
=== RUN   TestAccHerokuApp_ExternallySetBuildpacks
--- PASS: TestAccHerokuApp_ExternallySetBuildpacks (3.28s)
=== RUN   TestAccHerokuApp_Organization
--- PASS: TestAccHerokuApp_Organization (2.20s)
=== RUN   TestAccHerokuApp_Space
--- PASS: TestAccHerokuApp_Space (420.77s)
--- PASS: TestAccHerokuApp_EmptyConfigVars (1.94s)
PASS
ok      github.com/terraform-providers/terraform-provider-heroku/heroku 453.635s
```

### DIFF
--- a/heroku/import_heroku_app_test.go
+++ b/heroku/import_heroku_app_test.go
@@ -21,10 +21,9 @@ func TestAccHerokuApp_importBasic(t *testing.T) {
 				Config: testAccCheckHerokuAppConfig_basic(appName),
 			},
 			{
-				ResourceName:            "heroku_app.foobar",
-				ImportState:             true,
-				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"config_vars"},
+				ResourceName:      "heroku_app.foobar",
+				ImportState:       true,
+				ImportStateVerify: true,
 			},
 		},
 	})
@@ -48,10 +47,9 @@ func TestAccHerokuApp_importOrganization(t *testing.T) {
 				Config: testAccCheckHerokuAppConfig_organization(appName, org),
 			},
 			{
-				ResourceName:            "heroku_app.foobar",
-				ImportState:             true,
-				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"config_vars"},
+				ResourceName:      "heroku_app.foobar",
+				ImportState:       true,
+				ImportStateVerify: true,
 			},
 		},
 	})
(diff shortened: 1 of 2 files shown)


danp:
Related to https://github.com/hashicorp/terraform/issues/15962?

catsby:
@danp they appear to be unrelated; my change still gives that weird diff. The diff presented is... misleading? The net result is that var `FOO` is still there with that value.

radeksimko:
oh btw. keep in mind that we don't print out deprecation warnings for computed-only field, so this won't have any effect in the UI.

https://github.com/hashicorp/terraform/issues/7569

catsby:
I'm going to go ahead and merge this on Friday, November 3rd, if I don't hear otherwise before then. Thanks!

PR #20, opened 2017-10-27
resource_heroku_app: Add resourceHerokuAppExists

### LINKED ISSUE #4: provider/heroku: Error on refresh if App is missing
_This issue was originally opened by @catsby as hashicorp/terraform#14286. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Resource `heroku_app` seems to throw an error if the App is deleted out of band from Terraform:

- https://github.com/hashicorp/terraform/blob/master/builtin/providers/heroku/resource_heroku_app.go#L348

The `Read` method (`resourceHerokuAppRead`) should detect a specific "App no longer exists" type error, remove the app from state with `d.SetId("")`, and log the removal. 

Could also implement an `Exists` method, like so:

- https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_kms_key.go#L373

darkowlzz:
fixes #4

### DIFF
--- a/heroku/resource_heroku_app.go
+++ b/heroku/resource_heroku_app.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net/url"
 
 	"github.com/cyberdelia/heroku-go/v3"
 	multierror "github.com/hashicorp/go-multierror"
@@ -99,6 +100,7 @@ func resourceHerokuApp() *schema.Resource {
 		Read:   resourceHerokuAppRead,
 		Update: resourceHerokuAppUpdate,
 		Delete: resourceHerokuAppDelete,
+		Exists: resourceHerokuAppExists,
 
 		Importer: &schema.ResourceImporter{
 			State: resourceHerokuAppImport,
@@ -454,6 +456,26 @@ func resourceHerokuAppDelete(d *schema.ResourceData, meta interface{}) error {
 
 	d.SetId("")
 	return nil
+}
+
+func resourceHerokuAppExists(d *schema.ResourceData, meta interface{}) (bool, error) {
+	var err error
+	client := meta.(*heroku.Service)
+
+	if isOrganizationApp(d) {
+		_, err = client.OrganizationAppInfo(context.TODO(), d.Id())
+	} else {
+		_, err = client.AppInfo(context.TODO(), d.Id())
+	}
+	if err != nil {
+		// Make sure it's a missing app error.
+		if herr, ok := err.(*url.Error).Err.(heroku.Error); ok && herr.ID == "not_found" {
+			return false, nil
+		}
+		return false, err
+	}
+
+	return true, nil
 }
 
 func resourceHerokuAppRetrieve(id string, organization bool, client *heroku.Service) (*application, error) {
--- a/heroku/resource_heroku_app_test.go
+++ b/heroku/resource_heroku_app_test.go
@@ -40,6 +40,7 @@ func TestAccHerokuApp_Basic(t *testing.T) {
 func TestAccHerokuApp_Disappears(t *testing.T) {
 	var app heroku.App
 	appName := fmt.Sprintf("tftest-%s", acctest.RandString(10))
+	appStack := "cedar-14"
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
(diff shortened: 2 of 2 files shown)


heroku/resource_heroku_app.go
```diff
@@ -456,6 +458,29 @@ func resourceHerokuAppDelete(d *schema.ResourceData, meta interface{}) error {
...
+
+	app := application{Id: d.Id(), Client: client, Organization: organizationApp}
+	if organizationApp {
+		_, err = app.Client.OrganizationAppInfo(context.TODO(), app.Id)
+	} else {
+		_, err = app.Client.AppInfo(context.TODO(), app.Id)
+	}
+	if err != nil {
+		// Make sure it's a missing app error.
+		if strings.Contains(err.Error(), "Couldn't find that app") {
```
bernerdschaefer: Rather than looking at the string, it should be possible to example the structured [error](https://godoc.org/github.com/cyberdelia/heroku-go/v3#Error) returned by the client. In this case, it will return [not_found](https://devcenter.heroku.com/articles/platform-api-reference#client-error-responses).

I think something like this would work, though you may also need to unwrap the error from `*url.Error` first.

```go
if err != nil {
  if herr, ok := err.(*heroku.Error); ok && herr.ID == "not_found" {
    return false, nil
  }
  // ..
}
```

heroku/resource_heroku_app.go
```diff
@@ -456,6 +458,29 @@ func resourceHerokuAppDelete(d *schema.ResourceData, meta interface{}) error {
 	return nil
 }
 
+func resourceHerokuAppExists(d *schema.ResourceData, meta interface{}) (bool, error) {
+	var err error
+	client := meta.(*heroku.Service)
+
+	organizationApp := isOrganizationApp(d)
+
+	app := application{Id: d.Id(), Client: client, Organization: organizationApp}
```
bernerdschaefer: This doesn't appear to be used, except to access fields that are already available as values here.

I think it could be removed if the lines below are updated to be something like `_, err = client.OrganizationAppInfo(context.TODO(), d.Id())`.

heroku/resource_heroku_app.go
darkowlzz: Thank you for teaching this unwrapping trick 😊 I knew string comparison with error isn't the right thing to do but didn't knew the right way to handle it. ❤️

darkowlzz:
@bernerdschaefer Thanks for reviewing. I have added `TestAccHerokuApp_Disappears`, which deletes the app and runs a plan.
```
$ make testacc TESTARGS='-run=TestAccHerokuApp_Disappears'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccHerokuApp_Disappears -timeout 120m
?   	github.com/terraform-providers/terraform-provider-heroku	[no test files]
=== RUN   TestAccHerokuApp_Disappears
--- PASS: TestAccHerokuApp_Disappears (4.94s)
PASS
ok  	github.com/terraform-providers/terraform-provider-heroku/heroku	4.976s
```

And the error unwrapping changes have been squashed into the first commit.

tombuildsstuff:
Fixed the merge conflict and the tests pass:

```
$ acctests heroku TestAccHerokuApp_
=== RUN   TestAccHerokuApp_importBasic
--- PASS: TestAccHerokuApp_importBasic (4.65s)
=== RUN   TestAccHerokuApp_importOrganization
--- SKIP: TestAccHerokuApp_importOrganization (0.00s)
    import_heroku_app_test.go:41: HEROKU_ORGANIZATION is not set; skipping test.
=== RUN   TestAccHerokuApp_Basic
--- PASS: TestAccHerokuApp_Basic (3.52s)
=== RUN   TestAccHerokuApp_Disappears
--- PASS: TestAccHerokuApp_Disappears (1.77s)
=== RUN   TestAccHerokuApp_Change
--- PASS: TestAccHerokuApp_Change (6.66s)
=== RUN   TestAccHerokuApp_NukeVars
--- PASS: TestAccHerokuApp_NukeVars (6.44s)
=== RUN   TestAccHerokuApp_Buildpacks
--- PASS: TestAccHerokuApp_Buildpacks (9.59s)
=== RUN   TestAccHerokuApp_ExternallySetBuildpacks
--- PASS: TestAccHerokuApp_ExternallySetBuildpacks (5.41s)
=== RUN   TestAccHerokuApp_Organization
--- SKIP: TestAccHerokuApp_Organization (0.00s)
    resource_heroku_app_test.go:220: HEROKU_ORGANIZATION is not set; skipping test.
=== RUN   TestAccHerokuApp_Space
--- SKIP: TestAccHerokuApp_Space (0.00s)
    resource_heroku_app_test.go:247: HEROKU_ORGANIZATION is not set; skipping test.
--- PASS: TestAccHerokuApp_EmptyConfigVars (4.22s)
PASS
ok      github.com/terraform-providers/terraform-provider-heroku/heroku    42.298s
```

PR #28, opened 2017-11-21
Add trusted_ip_ranges attribute to space resource

vanstee:
Uses heroku's inbound ruleset api to set a list of trusted ip ranges per space, which maps directly to the form on the network tab for a space in the web ui.


bernerdschaefer:
Thanks @vanstee!

PR #32, opened 2017-12-18
Confirm add-on creation

dmathieu:
Some add-ons require passing the `confirm` attribute to be created.
For example, if we wish to provision a non-private plan within a private space app.


bernerdschaefer:
Thanks! Tests are all still green 👍 .

PR #35, opened 2017-12-20
Wait until new release has status succeeded when updating config vars

dmathieu:
Updating a config will create a new release.
If the app is using release phase, it will take a couple minutes before that release is available and used.

This change makes us wait before moving on.
So, for example, if the config var sets credentials and another resource requires those credentials to be set to be created, we shouldn't try creating that second resource until the release is successful.


dmathieu:
Ping @bernerdschaefer @justincampbell for review

Issue #30, opened 2017-11-22
Terraform tries to delete DATABASE_URL env var.


michal-januszkiewicz:
Hi,

I have a config with a database addon which generated `DATABASE_URL` after the first time I ran `terraform apply` and everything was ok, the `DATABASE_URL` was generated and added to heroku app but then I noticed I need to change some of the env vars so I edited the terraform file and after running `terraform plan` the output showed me it would change `DATABASE_URL` to an empty string.

Here is what I have in config vars
```hcl
  config_vars {
    HANAMI_ENV = "production",
    NODE_MODULES_CACHE = "false",
    NPM_CONFIG_PRODUCTION = "false",
  }
```
As you can see there is no `DATABASE_URL` specified since it is computed.

And here is the output of `terraform plan`
```
~ heroku_app.staging
      config_vars.0.%:                     "4" => "3"
      config_vars.0.DATABASE_URL:          "postgres://agextps..." => ""
      config_vars.0.NODE_MODULES_CACHE:    "0" => "false"
      config_vars.0.NPM_CONFIG_PRODUCTION: "0" => "false"
```

justincampbell:
Duplicate of #26, and being addressed in #25. In the meantime, you can pin the Heroku Terraform module to v0.1.0:

```hcl
provider "heroku" {
  version = "0.1.0"
}
```

Thanks! Sorry for the inconvenience.

justincampbell:
[v0.1.2 has been released](https://github.com/terraform-providers/terraform-provider-heroku/blob/master/CHANGELOG.md#012-january-04-2018) with a fix for this. Please test and report any issues. Thanks!

Issue #26, opened 2017-11-13
Heroku: DATABASE_URL overwriting


hashibot:
_This issue was originally opened by @JeremyLoy as hashicorp/terraform#16631. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

See [here](https://devcenter.heroku.com/changelog-items/438). Heroku added in 2014 that DATABASE_URL will get automatically set when added on

My Terraform File
```
resource "heroku_app" "default" {
  name = "terraformpgconfigbug"
  region = "us"
}

resource "heroku_addon" "db" {
  app  = "${heroku_app.default.name}"
  plan = "heroku-postgresql:hobby-dev"
}
```

### Steps to Reproduce
1. `terraform plan`
2. `terraform apply` -> should work perfectly fine
3. `heroku config` - > shows that DATABASE_URL is set automatically
4. `terraform plan` -> observe DATABASE_URL will be overwritten with "" which is incorrect

justincampbell:
@JeremyLoy I believe this was introduced in v0.1.1, and is being addressed in #25. You can pin your Terraform provider to v0.1.0 to avoid this issue for now, and then upgrade to the next release where this should be fixed.

```hcl
provider "heroku" {
  version = "0.1.0"
}
```

JeremyLoy:
Thanks @justincampbell. That PR looks great, glad the bot led me here.

JamesBelchamber:
Hey @justincampbell @radeksimko should we still wait on #25 or should there be a release that reverts the changes in #17 and fixes this for users of the provider?

I'm concerned that we're making perfect the enemy of good here :/

justincampbell:
Reverted in #36

PR #36, opened 2017-12-20
Revert changes to external config_vars deletion

justincampbell:
This reverts changes introduced in #17, which effectively removed `config_vars` which were set on an app externally via the CLI or UI. This had the unintended effect of removing variables added by addons or other desired services from Heroku.

Initially, we planned on adding an `ignore_vars` or `external_config_vars` attributes, which would be a list of config vars to not remove if they were noticed by Terraform. There's a WIP of that implemented here: https://github.com/terraform-providers/terraform-provider-heroku/commit/7223397

After experimenting with this, and discussing with @catsby, we've come to the conclusion that it would be better to just revert these changes. The `external_config_vars` attribute required writing to the state before taking effect, requiring awkward use of `ignore_changes` to cleanly ignore an already-added external config var.

Previous fix attempt which ignored variables from addons. and discussion: #25
Issues: #24, #26, #30

/cc @robertrossmann @JamesBelchamber @JeremyLoy

### DIFF
--- a/heroku/import_heroku_app_test.go
+++ b/heroku/import_heroku_app_test.go
@@ -22,9 +22,10 @@ func TestAccHerokuApp_importBasic(t *testing.T) {
 				Config: testAccCheckHerokuAppConfig_basic(appName, appStack),
 			},
 			{
-				ResourceName:      "heroku_app.foobar",
-				ImportState:       true,
-				ImportStateVerify: true,
+				ResourceName:            "heroku_app.foobar",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"config_vars"},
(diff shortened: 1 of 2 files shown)


JamesBelchamber:
I'd still really love to see a solution whereby we actually manage a `heroku_app` instead of just managing a corner of it, one of the benefits of Terraform is that it will detect and "heal" changes made manually.

That said, I support this revert :)

robertrossmann:
I support this revert - having terraform only manage vars explicitly defined in configuration is still better than crashing due to vars added by addons. 👍

PS. I assume there is no way of knowing beforehand which environment variables will be added to an app by an addon?

JeremyLoy:
LGTM 👍

PR #38, opened 2018-01-28
Adding Automated Certificate Management support

kubek2k:
* added ACM support - as an app field
* ACM is only usable in org apps - its due to Heroku constraints - thats why I've enabled it for any app - it could potentially work for an existing app that has dyno formation scaled to `Standard-1X`
* use the simple app retrieval for most fields, because now it seems to make sense (we're still missing the locked field though)

### DIFF
--- a/heroku/resource_heroku_app.go
+++ b/heroku/resource_heroku_app.go
@@ -42,43 +42,32 @@ func (a *application) Update() error {
 	var errs []error
 	var err error
 
-	if !a.Organization {
-		app, err := a.Client.AppInfo(context.TODO(), a.Id)
-		if err != nil {
-			errs = append(errs, err)
-		} else {
-			a.App = &herokuApplication{}
-			a.App.Name = app.Name
-			a.App.Region = app.Region.Name
-			a.App.Stack = app.BuildStack.Name
-			a.App.GitURL = app.GitURL
-			a.App.WebURL = app.WebURL
-			a.App.Acm = app.Acm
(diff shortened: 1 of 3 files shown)


justincampbell:
Awesome!

I wonder if having this as it's own resource (`heroku_acm` or `heroku_certs_auto`) would be a better representation of the APIs involved. It would also allow you to mark the lifecycle of ACM as dependent on domains being added.

Either way, excited for this to land!

kubek2k:
@justincampbell well - on one hand the modification endpoints are separate, on the read is right within the app. On the other hand you are right that it would be nice to be able to make acm dependant on the domain name assignment. While this assignment is external to the heroku app - there is nothing that stops the user to make a `depends_on` entry on the actual app.

justincampbell:
> there is nothing that stops the user to make a depends_on entry on the actual app

@kubek2k I don't think this is true, as the `heroku_domain` resource is typically dependent on a `heroku_app`, so creating a dependency the other way would be a cycle.

Sorry, I didn't realize the read was on the app itself. I suppose that makes it more difficult to have a separate resource.

kubek2k:
> ... so creating a dependency the other way would be a cycle.

you are right!

> Sorry, I didn't realize the read was on the app itself. I suppose that makes it more difficult to have a separate resource.

Well its not as hard - simple retrieval of app should do it. The issue is, I am not sure I am able to go through the pain of Go development again ;)

PR #41, opened 2018-02-12
Heroku app datasource

sheax0r:
Enables `heroku_app` data objects like this:

```
data "heroku_app" "my_app" {
  name = "my-app"
}
```

Has two tests:
* TestAccDatasourceHerokuApp_Basic (tests attributes common to all apps)
* TestAccDatasourceHerokuApp_Advanced (tests attributes only present in spaces apps)

I have run these tests, and I've also run `TestAccResourceHerokuApp_Basic`, since this involves a refactor of some `heroku_app` functionality.

### DIFF
--- a/heroku/data_source_heroku_app.go
+++ b/heroku/data_source_heroku_app.go
@@ -1,0 +1,120 @@
+package heroku
+
+import (
+	"time"
+
+	"github.com/cyberdelia/heroku-go/v3"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceHerokuApp() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceHerokuAppRead,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"space": {
+				Type:     schema.TypeString,
+				Computed: true,
+				Default:  nil,
+			},
+
+			"region": {
+				Type:     schema.TypeString,
+				Computed: true,
+				Default:  nil,
+			},
+
+			"stack": {
+				Type:     schema.TypeString,
+				Computed: true,
+				Default:  nil,
+			},
+
+			"buildpacks": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"config_vars": {
+				Type:     schema.TypeMap,
+				Computed: true,
+			},
+
+			"git_url": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"web_url": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"acm": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+
+			"heroku_hostname": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"organization": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"locked": {
+							Type:     schema.TypeBool,
+							Computed: true,
+						},
+
+						"personal": {
+							Type:     schema.TypeBool,
+							Computed: true,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func dataSourceHerokuAppRead(d *schema.ResourceData, m interface{}) error {
+	client := m.(*heroku.Service)
+
+	name := d.Get("name").(string)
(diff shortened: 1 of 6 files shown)


sigmavirus24:
This would be incredibly useful. 👍

ojacobson:
Extremely do want.

heroku/data_source_heroku_app.go
```diff
@@ -0,0 +1,120 @@
...
+		err := setOrganizationDetails(d, app)
+		if err != nil {
+			return err
+		}
+	}
+
+	setAppDetails(d, app)
+
+	d.Set("buildpacks", app.Buildpacks)
+	d.Set("config_vars", app.Vars)
```
joestump: Why not put these in `setAppDetails` as well?

heroku/data_source_heroku_app.go
```diff
@@ -0,0 +1,120 @@
...
+func dataSourceHerokuAppRead(d *schema.ResourceData, m interface{}) error {
+	client := m.(*heroku.Service)
+
+	name := d.Get("name").(string)
+	app, err := resourceHerokuAppRetrieve(name, true, client)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(time.Now().UTC().String())
```
joestump: I'm pretty sure data resources refresh on every run? I've noticed a lot of "run every time" data sources `SetId("-")` though I'm not entirely sure why. The docs say this:

> If the arguments of a data instance contain no references to computed values, such as attributes of resources that have not yet been created, then the data instance will be read and its state updated during Terraform's "refresh" phase, which by default runs prior to creating a plan. This ensures that the retrieved data is available for use during planning and the diff will show the real values obtained.

Less a statement on code and more a question on what's best practices here?

heroku/data_source_heroku_app.go
```diff
@@ -0,0 +1,120 @@
...
+
+	if app.Organization {
```
joestump: On compile, I got this error when I tried to do a similar check on a similar struct.

```
heroku/data_source_heroku_space.go:72:2: non-bool space.Organization (type struct { Name string "json:\"name\" url:\"name,key\"" }) used as if condition
```

heroku/data_source_heroku_app.go
sheax0r: Hey @joestump - I just went with a pattern for this that I saw used in some AWS datasources. I don't know what's the "best practice" - I just assumed that AWS knows what they're doing. I feel like it probably doesn't matter too much what we do here for datasources.

heroku/data_source_heroku_app.go
sheax0r: I don't put this in `setAppDetails` as well because that method is also invoked [here](https://github.com/sheax0r/terraform-provider-heroku/blob/735fc12a76e109234fb63fc9707c8e4d7bbe29ec/heroku/resource_heroku_app.go#L420), which doesn't have the same logic around setting buildpacks and config vars.

Config vars have some special logic in the `heroku_app` resource unlike the datasource, because it only deals with managed config vars, and ignores other ones. Whereas the datasource is simpler, and returns all config vars, regardless of what generated or if they're managed by the terraform scripts or not.

Buildpacks have similar concerns in the resource.

website/docs/d/app.html.markdown
```diff
@@ -0,0 +1,54 @@
...
+   for pointing DNS records.
+* `config_vars` - (Optional) A map of all of the configuration variables for the app.
+* `acm` - (Required) True if Heroku ACM is enabled for this app, false otherwise.
+* `organization` - (Optional) The organization that owns this app, if the app is owned by an organization. The fields for this block are documented below.
+
+The `organization` block supports:
+
+* `name` (string) - The name of the organization.
+* `locked` (boolean)
+* `personal` (boolean)
```
joestump: I think we'll need to add this to the sidebar in the ERB template as well. See the [`aws`](https://github.com/terraform-providers/terraform-provider-aws/blob/master/website/aws.erb#L24-L30) module's example here.

sheax0r: @joestump thanks, I didn't realize! 

Do you know how we can test this stuff? I made some modifications that I *think* will work in the sidebar, but I have no idea how I can actually verify it.

sheax0r:
@bernerdschaefer fyi: this has passed acceptance tests, except for [these](https://github.com/terraform-providers/terraform-provider-heroku/issues/49) which also fail on master, and it has been reviewed by @joestump. Can you merge for me? I could merge this myself, but I feel like it's probably best if I do not merge my own PRs (or at least, not unless another committer reviews them).

website/docs/d/app.html.markdown
joestump: I don’t know of a good way of testing this outside of compiling the docs and using Selenium, which sounds heavy. Maybe testing is handled downstream by HashiCorp?

bernerdschaefer:
Thanks for contributing this, @sheax0r!

PR #46, opened 2018-02-21
Retrieve the outbound IPs for a space

sigmavirus24:
Allow a user to introspect and deal with a space's outbound IP addresses
in the event they need to use them elsewhere in their automation.


joestump:
#63 added this functionality. Closing this out.

PR #51, opened 2018-03-14
Make trusted_ip_ranges a computed attribute.

joestump:
Fixes #42. When `Computed: true` is set:

```
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

heroku_space.test: Refreshing state... (ID: da77b7de-5a29-422d-8631-88504ac40134)

------------------------------------------------------------------------

No changes. Infrastructure is up-to-date.
```

When it is not set:

```
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

heroku_space.test: Refreshing state... (ID: da77b7de-5a29-422d-8631-88504ac40134)

------------------------------------------------------------------------

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  ~ heroku_space.test
      trusted_ip_ranges.#: "1" => "0"
      trusted_ip_ranges.0: "0.0.0.0/0" => ""


Plan: 0 to add, 1 to change, 0 to destroy.
```

PR #52, opened 2018-03-20
Document heroku_space trusted_ip_ranges, make required

Xe:
Heroku private spaces [require an IP address range of allowable addresses](https://devcenter.heroku.com/articles/private-spaces#trusted-ip-ranges). Currently the provider supports setting this, but it is not enforced or documented. This patch adds web documentation for this attribute of a `heroku_space` and makes it required, as a private space without any allowed IP address ranges is effectively useless.

### DIFF
--- a/heroku/resource_heroku_space.go
+++ b/heroku/resource_heroku_space.go
@@ -48,7 +48,7 @@ func resourceHerokuSpace() *schema.Resource {
 			"trusted_ip_ranges": {
 				Type:     schema.TypeList,
 				Computed: true,
-				Optional: true,
+				Optional: false,
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
 				},
--- a/website/docs/r/space.html.markdown
+++ b/website/docs/r/space.html.markdown
@@ -37,6 +37,7 @@ The following arguments are supported:
 * `name` - (Required) The name of the space.
 * `organization` - (Required) The name of the organization which will own the space.
 * `region` - (Optional) The region that the space should be created in.
+* `trusted_ip_ranges` - (Required) The list of IP address CIDR ranges (0.0.0.0/0, etc) that can connect to apps hosted in this private space.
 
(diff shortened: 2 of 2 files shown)


bernerdschaefer:
+1 to documenting, -1 to making it required.

When https://github.com/terraform-providers/terraform-provider-heroku/pull/51 is released, terraform will no longer remove the default trusted_ip_ranges value if you don't specify it.

heroku/resource_heroku_space.go
```diff
@@ -48,7 +48,7 @@ func resourceHerokuSpace() *schema.Resource {
 			"trusted_ip_ranges": {
 				Type:     schema.TypeList,
 				Computed: true,
-				Optional: true,
+				Optional: false,
```
joestump: Let's back this out in favor of #51.

website/docs/r/space.html.markdown
```diff
@@ -37,6 +37,7 @@ The following arguments are supported:
 * `name` - (Required) The name of the space.
 * `organization` - (Required) The name of the organization which will own the space.
 * `region` - (Optional) The region that the space should be created in.
+* `trusted_ip_ranges` - (Required) The list of IP address CIDR ranges (0.0.0.0/0, etc) that can connect to apps hosted in this private space.
```
joestump: Change `Required` to `Optional`. I'd mention that Heroku defaults this to `["0.0.0.0/0"]`.

joestump:
We've recently released the [`heroku_space_inbound_ruleset`](https://www.terraform.io/docs/providers/heroku/r/space_inbound_ruleset.html) for managing Heroku Private Space IP filtering. As part of this change we've deprecated the `trusted_ip_ranges` attribute on the `heroku_space` resource. Given these two developments, I'm going to mark this as closed and pour one out for our friend `trusted_ip_ranges`.

PR #53, opened 2018-04-10
WIP: Make space use an empty list for the default value of trusted_ip_ranges.

sheax0r:
If trusted_ip_ranges isn't set, Heroku defaults to [0.0.0.0]. However, terraform
itself gets the value in the state set to [], resulting in a newly created space's state being
out of sync.

It is not possible (AFAICT at this point) to specify a default value for a List. Thefore,
we have no way to specify that the state should default to [0.0.0.0] that terraform can understand.

This change will break backwards compatibility - we might want to consider also making the
trusted_ip_range a required attribute, since a default of [] might be confusing to users.

### DIFF
--- a/heroku/resource_heroku_space.go
+++ b/heroku/resource_heroku_space.go
@@ -89,26 +89,32 @@ func resourceHerokuSpaceCreate(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("Error waiting for Space (%s) to become available: %s", d.Id(), err)
 	}
 
-	if ranges, ok := d.GetOk("trusted_ip_ranges"); ok {
-		var rules []*struct {
+	// If there are no ranges specified, default to empty list of ranges
+	ranges, ok := d.GetOk("trusted_ip_ranges")
+	if !ok {
+		ranges = make([]interface{}, 0)
+	}
+
+	log.Printf("[DEBUG] Setting trusted_ip_ranges")
+	var rules []*struct {
+		Action string `json:"action" url:"action,key"`
+		Source string `json:"source" url:"source,key"`
+	}
+	for _, r := range ranges.([]interface{}) {
+		log.Printf("Setting range to : %s", r.(string))
+		rules = append(rules, &struct {
 			Action string `json:"action" url:"action,key"`
 			Source string `json:"source" url:"source,key"`
-		}
-		for _, r := range ranges.([]interface{}) {
-			rules = append(rules, &struct {
-				Action string `json:"action" url:"action,key"`
-				Source string `json:"source" url:"source,key"`
-			}{
-				Action: "allow",
-				Source: r.(string),
-			})
-		}
+		}{
+			Action: "allow",
+			Source: r.(string),
+		})
+	}
 
-		opts := heroku.InboundRulesetCreateOpts{Rules: &rules}
-		_, err := client.InboundRulesetCreate(context.TODO(), space.ID, opts)
-		if err != nil {
-			return fmt.Errorf("Error creating Trusted IP Ranges for Space (%s): %s", space.ID, err)
-		}
+	ruleOpts := heroku.InboundRulesetCreateOpts{Rules: &rules}
+	_, ruleErr := client.InboundRulesetCreate(context.TODO(), space.ID, ruleOpts)
+	if ruleErr != nil {
+		return fmt.Errorf("Error creating Trusted IP Ranges for Space (%s): %s", space.ID, ruleErr)
 	}
 
 	return resourceHerokuSpaceRead(d, meta)


sheax0r:
@Xe @bernerdschaefer I'm currently running the complete test suite using this, but it at least fixes `TestAccHerokuApp_Space`.

It will break backwards compatibility, but I haven't been able to figure out a better way to reconcile Heroku's "0.0.0.0" default with Terraform's interpretation of "no default specified" as "[]".

What do you folks think of this? We'll probably want to bump a significant version number if we add this change.

sheax0r:
At this point, `TestAccHerokuSpace_Basic` still fails. I'll try to look into that tomorrow if I have time.

```
--- FAIL: TestAccHerokuSpace_Basic (443.69s)
	testing.go:434: Step 1 error: After applying this step, the plan was not empty:

		DIFF:

		UPDATE: heroku_space.foobar
		  trusted_ip_ranges.#: "0" => "2"
		  trusted_ip_ranges.0: "" => "8.8.8.8/32"
		  trusted_ip_ranges.1: "" => "8.8.8.0/24"

		STATE:

		heroku_space.foobar:
		  ID = 630580e8-ebbe-478f-badd-7849278a7325
		  name = tftest-xs49uq64sp
		  organization = <REDACTED>
		  region = virginia
		  trusted_ip_ranges.# = 0
```

joestump:
@sheax0r have you tried setting `trusted_ip_ranges` to `Computed: true`?

sheax0r:
@joestump I definitely did try that, and it didn't just solve the problems.

joestump:
@sheax0r so my understanding is that after a space is created the [update function calls `resourceHerokuSpaceRead `](https://github.com/terraform-providers/terraform-provider-heroku/blob/master/heroku/resource_heroku_space.go#L115), which in turn calls [`SpaceStateRefreshFunc `](https://github.com/terraform-providers/terraform-provider-heroku/blob/master/heroku/resource_heroku_space.go#L121), which in turn [sets `TrustedIPRanges`](https://github.com/terraform-providers/terraform-provider-heroku/blob/master/heroku/resource_heroku_space.go#L212-L219). 

In other words, no matter what Heroku defaults to the `trusted_ip_ranges` should be set to default on create.

joestump:
The odd thing about this failing test is that things work fine when you emulate the test from the CLI:

**1: Create a test.tf**

```hcl
resource "heroku_space" "trusted-ip-test" {
  name         = "trusted-ip-bugs-test"
  organization = "sfdc-ohana-terraform-sandbox"
  region       = "virginia"

  trusted_ip_ranges = [
    "8.8.8.8/32",
    "8.8.8.0/24",
  ]
}
```

**2. Run `plan` and `apply`**

Verified that the IP ranges are set correctly in Heroku:

screen shot 2018-06-01 at 12 20 24 pm

**3. Check state file**

```json
{
    "version": 3,
    "terraform_version": "0.11.7",
    "serial": 1,
    "lineage": "XXXXXX-XXXX-XXXX-XXX-XXXXX",
    "modules": [
        {
            "path": [
                "root"
            ],
            "outputs": {},
            "resources": {
                "heroku_space.trusted-ip-test": {
                    "type": "heroku_space",
                    "depends_on": [],
                    "primary": {
                        "id": "XXXXXXX-XXXXX-XXXX-XXXX-XXXXXXXXX",
                        "attributes": {
                            "id": "XXXXXXX-XXXXX-XXXX-XXXX-XXXXXXXXX",
                            "name": "trusted-ip-bugs-test",
                            "organization": "my-terraform-sandbox",
                            "region": "virginia",
                            "trusted_ip_ranges.#": "2",
                            "trusted_ip_ranges.0": "8.8.8.8/32",
                            "trusted_ip_ranges.1": "8.8.8.0/24"
                        },
                        "meta": {},
                        "tainted": false
                    },
                    "deposed": [],
                    "provider": "provider.heroku"
                }
            },
            "depends_on": []
        }
    ]
}
```

**4. Change the `name` of the space and run `plan`**

```
terraform plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

heroku_space.trusted-ip-test: Refreshing state... (ID: XXXXXXX-XXXXX-XXXX-XXXX-XXXXXXXXX)

------------------------------------------------------------------------

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  ~ heroku_space.trusted-ip-test
      name: "trusted-ip-bugs-test" => "trusted-ip-bugs-test-2"


Plan: 0 to add, 1 to change, 0 to destroy.
```

So my question is what would cause this work when you actually do the thing, but fail when being ran from the acceptance tests?

heroku/resource_heroku_space.go
```diff
@@ -90,26 +90,32 @@ func resourceHerokuSpaceCreate(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("Error waiting for Space (%s) to become available: %s", d.Id(), err)
 	}
 
-	if ranges, ok := d.GetOk("trusted_ip_ranges"); ok {
-		var rules []*struct {
+	// If there are no ranges specified, default to empty list of ranges
+	ranges, ok := d.GetOk("trusted_ip_ranges")
+	if !ok {
+		ranges = make([]interface{}, 0)
```
joestump: I think if you changed this to `[]string{"0.0.0.0/0"}` this specific issue might get fixed.

joestump:
Tests are all passing in #61. Why? `:asciishrug:`

joestump:
#63 merged in fixes to the acceptance tests. Closing this.

PR #59, opened 2018-06-01
Update the vendor directory with govend

joestump:
I'm not well versed in Go's vendor/package management strategies so it wouldn't surprise me if this is wrong. I ran:

* `govend -u github.com/cyberdelia/heroku-go`
* `govend github.com/hashicorp/go-multierror`
* `govend github.com/satori/uuid`


joestump:
Closing this out in favor of #60.

PR #83, opened 2018-06-28
Create a resource for managing space access

talbright:
TODO

- [x] documentation for HCL
- [x] Create
- [x] Read
- [x] Update
- [x] Delete
- [x] Get fix for Heroku API merged (heroku/api#9602)
- [x] Get update to `heroku-go` merged (cyberdelia/heroku-go/pull/27) 
- [x] Code cleanup
- [x] Integration testing
- [x] Acceptance tests


talbright:
@joestump No code yet, requesting feedback on the HCL

davidji99:
Don't forget to add the new resource to https://github.com/terraform-providers/terraform-provider-heroku/blob/master/website/heroku.erb as well.

joestump:
Two things come to mind:

* Should we call this `heroku_space_collaborator` to align with the [`heroku_team_collaborator`](https://www.terraform.io/docs/providers/heroku/r/team_collaborator.html) resource? 
* There's no delete endpoint for space access? How do I remove a collaborator from a space?

website/docs/r/space_access.markdown
```diff
@@ -0,0 +1,45 @@
...
+sidebar_current: "docs-heroku-resource-space-access"
+description: |-
+  Provides a Heroku Space access resource for managing access to the space. Note that members with admin privileges will
+  only have full access to a Heroku Space.
+---
+
+# heroku\_space\_access
+
+Provides a Heroku Space access resource for managing team member permissions to a space. Currently the only supported permission 
+is `app_creation`. Note that team admins will already have full access to a Heroku Space.
```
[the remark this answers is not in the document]
talbright: Agreed, I will fix.

talbright:
The terminology is a bit tricky. A `collaborator` has a different meaning then a `member`. A `collaborator` < `member`, and is intended to be used as someone you grant explicit access to apps outside your org. As a team `member` you can take on the roll of **(admin | member | viewer)** (Yay, a member-member). Keep in mind this is access to the space, app access is a separate matter and resource.

In this case, the API endpoint for this is a `member` resource on a `space` resource:

`https://api.heroku.com/spaces/{my-space}/members/{email@foo.com}`

References:
- https://devcenter.heroku.com/articles/app-privileges-in-heroku-organizations
- https://devcenter.heroku.com/articles/org-users-access

All this is my long way of saying I lean towards `member` terminology unless we are talking about a literal org `collaborator` (which we aren't in this case.)

The problem with `heroku_space_member` is it implies we are adding this member to the space, when in fact they are already by default a part of the space, but with no permissions (unless they have the admin role.)

Maybe `heroku_space_member_permissions` if you want to match the JSON or `heroku_space_member_access` if you want to match the UI?

talbright:
To answer your other question @joestump, you can't remove a member from a space, only restrict/remove permissions.

davidji99:
What about just `heroku_space_access`? But I guess `heroku_space_member_access` is more specific and clear on the resource's purpose.

As for the `DELETE` command, I guess we could remove the resource from state and do a `PATCH` to remove the `create_apps` perm?

talbright:
@davidji99 I was thinking along the same lines as you with regards to the `DELETE` behavior.  From a security point of view its the best for customers too.

talbright:
@joestump there's a bug with the Heroku API JSON schema, any suggestions on who I can talk to over at Heroku?

talbright:
@bernerdschaefer any thoughts on what this resource should be named? Does `heroku_space_access` work?

bernerdschaefer:
`heroku_space_member` sounds like a great resource name to me, though I also don't love departing from the platform schema names for things.

heroku/resource_heroku_space_app_access.go
```diff
@@ -0,0 +1,129 @@
...
+			},
+		},
+	}
+}
+
+//callback for schema Resource.Create
+//There's no actual method to create a space member, so we just need to import
+//the existing member into the state and update if needed.
+func resourceHerokuSpaceAppAccessCreate(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[DEBUG] resourceHerokuSpaceAppAccessCreate")
```
joestump: I'd add more debug to this output. There's no way for me to tell in the log output what access is being added. TF output is threaded and can have multiple instances of a single resource.

heroku/resource_heroku_space_app_access.go
```diff
@@ -0,0 +1,129 @@
...
+	err := resourceHerokuSpaceAppAccessUpdate(d, meta)
+	if err != nil {
+		return err
+	}
+	return resourceHerokuSpaceAppAccessRead(d, meta)
+}
```
joestump: I would remove this function entirely and use the same function for both create and delete. See the `heroku_space_inbound_ruleset` for an example.

website/docs/r/space_app_access.markdown
```diff
@@ -0,0 +1,45 @@
...
+sidebar_current: "docs-heroku-resource-space-app-access"
+description: |-
+  Provides a Heroku Space resource for managing app permissions for the entire space. Note that members with the admin role will
+  always have full permissions to a Heroku Space. The provided member must already exist in your Heroku organization.
+---
+
+# heroku\_space\_member
+
+Provides a Heroku Space resource for managing app permissions for the entire space. Note that members with the admin role will
+always have full permissions to a Heroku Space. The provided member must already exist in your Heroku organization. The provided member must already exist in your Heroku organization. Currently the only supported permission is `create_apps`.
```
joestump: I'd move this note into `~> **NOTE:**` so that it's more clearly highlighted. See [this example](https://www.terraform.io/docs/providers/heroku/r/team_collaborator.html) of what it looks like.

website/docs/r/space_app_access.markdown
```diff
@@ -0,0 +1,45 @@
...
+  space = "${heroku_space.default.name}"
+  email = "member1@foobar.com"
+  permissions = ["create_apps"]
+}
+
+// Remove an existing team members permissions to the space
+resource "heroku_space_app_access" "member2" {
+  space = "${heroku_space.default.name}"
+  email = "member2@foobar.com"
+}
```
joestump: I don't think this should be documented like this. It's not very Terraform'y in nature. If someone had an existing space with existing members, I'd expect the workflow to go like this:

* Create the HCL for the space and membership access.
* Import the space and membership access into the state file.
* Delete the entire `heroku_space_app_access` resource from the HCL.
* Run TF and it removes all permissions and nukes data from state.

website/docs/r/space_app_access.markdown
```diff
@@ -0,0 +1,45 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `space` - (Required) The name of the space.
+* `email` - (Required) The email of the team member to set permissions for.
+* `permissions` - (Optional) The permissions to grant the team member for the space. Currently `create_apps` is the only supported permission. If not provided the member will have no permissions to the space. Members with admin role will always have `create_apps` permissions, which cannot be removed.
```
joestump: I don't think this should be optional. That or we should add a `Default: []string{"create_apps"}` to the attribute. And is there really only this permission? The API docs don't appear to have any permissions outlined.

heroku/resource_heroku_space_app_access.go
talbright: This just calls `Update`, where there's more detailed logging (what permissions are being changed and for whom.) Let me know if you think it needs more. Also, I fixed things so we now have logging of the HTTP request/responses, which is pretty helpful.

heroku/resource_heroku_space_app_access.go
talbright: There's some subtle differences between the two, so they don't share the exact same code. For this particular resource re-using the `Read` and `Update` callbacks seemed to work out the best.

website/docs/r/space_app_access.markdown
talbright: I might be missing what you mean, this is how it works AFAIK (but I might have missed something let me know.) Maybe the confusion comes from the fact that: leaving a `heroku_space_app_access` resource wo a `permissions` block in fact makes sure that user is granted no permissions.

website/docs/r/space_app_access.markdown
talbright: The way it works now, no permissions block means no permissions, that's why its optional. The concern I have about `create_apps` as the default, is its to permissive. I think a safer default is to deny.

Indeed, `create_apps` appears to be the only permission, I know its a bit odd. Having worked with the API a bit now I'm assuming there will be additional permissions added in the future. For example, why can't I grant the same permissions that can be granted to an individual app at the space level?

Overall I found it a bit tricky to get the behavior I was expecting from Terraform to marry well with the API. Given this additional info, let me know what you think.

heroku/resource_heroku_space_app_access.go
joestump: I would remove this log entirely then. It will be of no use in its current form once there are two of these resources in a file. There's no identifying information in it.

heroku/resource_heroku_space_app_access.go
joestump: Can you point out the differences? Near as I can tell the create code just calls update and returns the read. I'd just replace the update's `return nil` with `return resourceHerokuSpaceAppAccessRead(d, meta)`.

website/docs/r/space_app_access.markdown
joestump: When you "remove" something from the API you're generally also removing HCL from a config somewhere/somehow. Perhaps the comment above this is what's throwing me off? What I'm trying to ask is if you're suggesting people should never remove a `heroku_space_app_access` resource from HCL ever-ever or if this is supposed to be an example that denies access to a given space for a given user? I think you're saying it's the latter, which if that's the case we should tighten up this language to be clearer that you're not "removing" the person so much as "denying" them access to apps on that space.

heroku/resource_heroku_space_app_access.go
talbright: Yeah that's been removed, I just haven't pushed it up. That was left over from some tracing I was doing.

heroku/resource_heroku_space_app_access.go
talbright: https://github.com/talbright/terraform-provider-heroku/blob/3de7445d9c679af40e3159eafbebbd0def559c9a/heroku/resource_heroku_space_app_access.go#L96

website/docs/r/space_app_access.markdown
joestump: Hmmm, I still think it should be required, but I see why we'd not want to set a default. Going back to my confusion over your "remove" example, I think having `permissions = []` gives the "deny access" use case more clarity.

heroku/resource_heroku_space_app_access.go
joestump: I'm not commenting on the delete function. I'm saying there's no reason to have a separate _create_ function because near as I can tell it just calls update then read. I'd just replace `return nil` like outlined above.

davidji99:
Can you add an import test for this new resource?

website/docs/r/space_app_access.markdown
davidji99: This is confusing to me as well. Given that there is only one permission available for this resource right now, if I want to remove this permission for a user, I expect to just delete the HCL.

website/docs/r/space_app_access.markdown
davidji99: Maybe, this resource could be renamed as `heroku_space_member_permissions` or just `heroku_space_permissions`. I suggest this because it seems space access to a member is automatically granted if they are members of the organization? The ability to create apps in the space is not granted by default and collaborators do not have any access at all. 

So, I don't think we need to manage the 'access' or 'membership' the space, but only the permissions.

website/docs/r/space_app_access.markdown
talbright: Right...its the latter, removing them from the HCL doesn't remove them from a space, the Heroku API doesn't work that way. @davidji99 and I had some conversation around the correct behavior on removal, and felt that removing all permissions was the right way to go.

website/docs/r/space_app_access.markdown
talbright: I know we've all gone back and forth on the naming topic, it seems particularly hard to get this one right. I decided to fallback on following the conventions of the API (which is SpaceAppAccess.) Ultimately we'll go back and forth on this and I'll go crazy renaming the things. I call naming it `heroku_space_app_access` since that reflects at this time what the resource is doing. However, @davidji99 make the final call if you like something else I'll change it and everyone else needs to be happy with it 😄 

To answer your other question, by default all users exist in the space, so the API returns them, its just that they have no permissions.

davidji99: Ah yeah, follow the API naming convention. Scratch my suggestions.

talbright: @joestump lets go with making it required and if its empty, that would remove any permissions the user has:

talbright:
Is this more what you had in mind @joestump ?

talbright:
@davidji99 when you get the chance can you take a look?

davidji99:
👀 ing

website/docs/r/space_app_access.markdown
```diff
@@ -0,0 +1,45 @@
+---
+layout: "heroku"
+page_title: "Heroku: heroku_space_app_access"
+sidebar_current: "docs-heroku-resource-space-app-access"
+description: |-
+  Provides a Heroku Space resource for managing app permissions for the entire space. Members with the admin role will
```
davidji99: So really, this resource is for non-admins. I wonder what happens when try to use this new resource on admin?

Do we need to add a disclaimer here saying this is really just for non-admins? Something like "you don't need to use this resource on admins"?

talbright: I wondered the same thing and tried it out while testing things locally. No errors occur, but of course as expected the admin permissions don't actually change.

As far as the disclaimer, I was trying to communicate that above with the following statement:

> Members with the admin role will always have full permissions to a Heroku Space. 

Perhaps that should be changed to:

> Members with the admin role will always have full permissions to a Heroku space, so using this resource on an admin will have no affect

?

davidji99: yeah +1 for the latter

PR #84, opened 2018-06-28
Addon attachments are always re-created

### LINKED ISSUE #73: Addon attachments are always re-created
### Terraform Version
0.11.7

### Affected Resource(s)
Please list the resources as a list, for example:
- heroku_addon_attachment

```hcl
heroku_app "app1" {
  ...
}

heroku_addon "addon1" {
  app = "${heroku_app.app1.name}"
  ...
}

heroku_app "app2" {
  ...
}

heroku_addon_attachment "addon1-app2" {
  app_id = "${heroku_app.app2.id}"
  addon_id = "${heroku_addon.addon1.id}"
}
```

### Expected Behavior
Once applied, re-planning should require no changes

### Actual Behavior
Plan includes:
```
-/+ heroku_addon_attachment.addon1-app2 (new resource required)
      id:       "[UUID 1 removed]" => <computed> (forces new resource)
      addon_id: "[UUID 2 removed]" => "[addon1 name removed]" (forces new resource)
      app_id:   "app2" => "app2"
      name:     "MY_NAME" => "MY_NAME"
```

### Steps to Reproduce
1. `terraform apply`
2. `terraform plan`

### Important Factoids
I think this is because the `id` attribute of the `heroku_addon` is its name, not its UUID. Therefore, when interpolating the HCL, terraform thinks the `heroku_addon_attachment` should be between `app2` and `addon1.name`, but in the state it's between `app2` and `addon1's UUID which is not in the state at all`. Therefore, every plan requires re-creating the addon attachment.

### References
I think this was introduced by #39 .

davidji99:
Resolves #73

### DIFF
--- a/heroku/resource_heroku_addon.go
+++ b/heroku/resource_heroku_addon.go
@@ -31,6 +31,9 @@ func resourceHerokuAddon() *schema.Resource {
 			State: schema.ImportStatePassthrough,
 		},
 
+		SchemaVersion: 1,
+		MigrateState:  resourceHerokuAddonMigrate,
+
 		Schema: map[string]*schema.Schema{
 			"app": {
 				Type:     schema.TypeString,
--- a/heroku/resource_heroku_addon_attachment_migrate.go
+++ b/heroku/resource_heroku_addon_attachment_migrate.go
@@ -1,0 +1,51 @@
+package heroku
+
+import (
+	"context"
+	"fmt"
+	"github.com/cyberdelia/heroku-go/v3"
+	"github.com/hashicorp/terraform/terraform"
+	"log"
+)
+
+func resourceHerokuAddonAttachmentMigrateState(v int, is *terraform.InstanceState, meta interface{}) (*terraform.InstanceState, error) {
+	conn := meta.(*heroku.Service)
+
+	log.Printf("[DEBUG] Current version of state file is: v%v", v)
+
+	switch v {
+	case 0:
+		log.Println("[INFO] Found Heroku Addon Attachment state v0; migrating to v1")
+		return migrateAddonAttachmentStateV0toV1(is, conn)
+	default:
+		return is, fmt.Errorf("Unexpected schema version: %d", v)
+	}
+}
+
+// Migrate attachment addons in state file to use the UUID for addon_id instead of the NAME for addon_id
+func migrateAddonAttachmentStateV0toV1(is *terraform.InstanceState, client *heroku.Service) (*terraform.InstanceState, error) {
+	if is.Empty() || is.Attributes == nil {
+		log.Println("[DEBUG] Empty Heroku Addon Attachment State; nothing to migrate.")
+		return is, nil
+	}
+
+	log.Printf("[DEBUG] Attributes before migration: %#v", is.Attributes)
+
+	attachmentAppId := is.Attributes["app_id"]
+	attachmentAddOnId := is.Attributes["id"]
+
+	addon, err := client.AddOnInfoByApp(context.TODO(), attachmentAppId, attachmentAddOnId)
+	if err != nil {
+		return nil, fmt.Errorf("Error retrieving addon: %s", err)
+	}
+
+	addonId := addon.ID
+	if attachmentAddOnId != addonId {
+		log.Printf("[DEBUG] Setting attachment's addon_id to %s", addonId)
+		is.Attributes["addon_id"] = addonId
+	}
+
+	log.Printf("[DEBUG] Attributes after migration: %#v", is.Attributes)
+
+	return is, nil
+}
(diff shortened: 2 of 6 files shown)


heroku/resource_heroku_addon.go
```diff
@@ -143,6 +143,7 @@ func resourceHerokuAddonRead(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
+	d.Set("id", addon.ID)
```
bernerdschaefer: Can you explain why this would be needed? I'd think that https://github.com/terraform-providers/terraform-provider-heroku/pull/84/files#diff-1a1b7e50595b433f6c7e741170d655b8R105 in create would be handling this already.

davidji99: @bernerdschaefer ah i'm still messing with this PR's code so was trying some stuff out. but yeah, it's not needed.

davidji99:
I've changed the scope of this PR to add two migrations for `heroku_addon` and `heroku_addon_attachment` to migrate any `addon` ids that are using the NAME instead of the UUID. It is my conclusion that the intended (maybe correct) behavior is to use the addon's UUID as the ID since that's what is stored in state via this line: https://github.com/terraform-providers/terraform-provider-heroku/blob/master/heroku/resource_heroku_addon.go#L105

An example `terraform plan` output in debug mode on `heroku_addon` would look like this:

```
2018-07-10T13:00:06.863-0700 [DEBUG] plugin.terraform-provider-heroku: 2018/07/10 13:00:06 [DEBUG] Addon Id before migration: "<$ADDON_NAME>"
2018-07-10T13:00:06.863-0700 [DEBUG] plugin.terraform-provider-heroku: 2018/07/10 13:00:06 [DEBUG] Addon Attributes before migration: map[string]string{"config_vars.#":"2", "config_vars.0":"<REDACTED>", "config_vars.1":"<REDACTED>", "id":"<$ADDON_NAME>", "name":"<REDACTED>", "plan":"<REDACTED>", "provider_id":"<REDACTED>", "app":"<REDACTED>"}

2018-07-10T13:00:07.166-0700 [DEBUG] plugin.terraform-provider-heroku: 2018/07/10 13:00:07 [DEBUG] Addon Id after migration: "<$SOME_UUID>"
2018-07-10T13:00:07.166-0700 [DEBUG] plugin.terraform-provider-heroku: 2018/07/10 13:00:07 [DEBUG] Addon Attributes after migration: map[string]string{"config_vars.#":"2", "config_vars.0":"<REDACTED>", "config_vars.1":"<REDACTED>", "id":"<$SOME_UUID>", "name":"<REDACTED>", "plan":"<REDACTED>", "provider_id":"<REDACTED>", "app":"<REDACTED>"}
```

davidji99:
@bernerdschaefer @joestump what do you think of these two migration files?

heroku/resource_heroku_addon_attachment_migrate.go
```diff
@@ -0,0 +1,51 @@
...
+	conn := meta.(*heroku.Service)
+
+	log.Printf("[DEBUG] Current version of state file is: v%v", v)
+
+	switch v {
+	case 0:
+		log.Println("[INFO] Found Heroku Addon Attachment state v0; migrating to v1")
+		return migrateAddonAttachmentStateV0toV1(is, conn)
+	default:
+		return is, fmt.Errorf("Unexpected schema version: %d", v)
```
joestump: What happens after the migrate? Does TF track this? Or will `1` cause this to throw an error?

joestump:
These look reasonable. I just have that question about how TF tracks these migrations. Do we create migrations like this for each schema version? Or does that single entrypoint mux migrations moving forward?

heroku/resource_heroku_addon_attachment_migrate.go
davidji99: Hmm let me ask hashicorp about this - I was wondering the same thing.

davidji99:
In reply to Joe's question above, here's what I've understand from talking to Hashicorp:
- Only roll forward, no rollbacks
- Looking at current examples (AWS), the `switch` statement style is popular. However if you have multiple state migrations, you need to add previous migrate `func` to the prior `case` statement. See https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_route53_record_migrate.go#L20
- if your state version is too new then Terraform will, I believe, tell the user to upgrade the provider
- upgrades can be lossy, so downgrading isn't possible in the general case, though might work in some cases
- It can be separate `_migrate.go` file for each migration of a resource X, but it seems current examples keep them altogether.
- The migration runs on state fresh after an `apply`.

heroku/resource_heroku_addon.go
davidji99: took this out already. Fyi @bernerdschaefer

PR #86, opened 2018-06-29
WIP: Migrate from deprecated organizations to teams

joestump:
Heroku has [deprecated organizations](https://devcenter.heroku.com/changelog-items/1132) in favor of teams. Heroku Private Spaces are only supported on "team apps". Because `ForceNew: true` is set on the `organization` attribute, and there are two separate API endpoints/code paths, this PR is introducing a preference for the `team` attribute while falling back to `organization`.

People using the provider wishing to use `team` instead of `organization` would need to do state manipulations to port existing `heroku_app` resources over.

### DIFF
--- a/heroku/data_source_heroku_app.go
+++ b/heroku/data_source_heroku_app.go
@@ -68,8 +68,10 @@ func dataSourceHerokuApp() *schema.Resource {
 			},
 
 			"organization": {
-				Type:     schema.TypeList,
-				Computed: true,
+				Type:          schema.TypeMap,
+				Computed:      true,
+				ConflictsWith: []string{"team"},
+				Deprecated:    "Heroku has deprecated organizations. Use team instead.",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"name": {
@@ -85,6 +87,32 @@ func dataSourceHerokuApp() *schema.Resource {
 						"personal": {
 							Type:     schema.TypeBool,
 							Computed: true,
+						},
+					},
+				},
+			},
+
+			"team": {
+				Type:          schema.TypeMap,
+				Computed:      true,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"organization"},
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+
+						"locked": {
+							Type:     schema.TypeBool,
+							Optional: true,
+						},
+
+						"personal": {
+							Type:     schema.TypeBool,
+							Optional: true,
 						},
 					},
 				},
--- a/heroku/data_source_heroku_app.go
+++ b/heroku/data_source_heroku_app.go
@@ -104,8 +104,8 @@ func dataSourceHerokuAppRead(d *schema.ResourceData, m interface{}) error {
 
 	d.SetId(time.Now().UTC().String())
 
-	if app.Organization {
-		err := setOrganizationDetails(d, app)
+	if app.Team {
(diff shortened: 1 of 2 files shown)


heroku/resource_heroku_app.go
```diff
@@ -170,9 +172,38 @@ func resourceHerokuApp() *schema.Resource {
...
+						"personal": {
+							Type:     schema.TypeBool,
+							Optional: true,
+						},
+					},
+				},
+			},
+
+			"team": {
+				Type:          schema.TypeList,
```
bernerdschaefer: Do you know the motivation for making this a list and not a map?

bernerdschaefer:
Looking good!

Two things I'm thinking about for this:

* Might it make sense to split the user-facing changes out from the internal ones? We should be able to switch to the team app endpoints and internal data structures (and start supporting new features) without alerting the schema.

* Is there prior art, tooling, or support available in the maintainer Slack for doing a graceful field transition?

heroku/resource_heroku_app.go
joestump: I just copied and pasted the `organization` definition. I don't think we need it to be `TypeList` unless multiple stanzas are allowed (e.g. `stage` in `aws_codepipeline`).

joestump:
@bernerdschaefer there's a facility for doing schema migrations internally on the fly. @davidji99 and I have it on our list of things to take a look at.

heroku/data_source_heroku_app.go
```diff
@@ -68,8 +68,10 @@ func dataSourceHerokuApp() *schema.Resource {
 			},
 
 			"organization": {
-				Type:     schema.TypeList,
-				Computed: true,
+				Type:          schema.TypeMap,
```
bernerdschaefer: Do we need to keep org as a list to maintain backwards compatibility with existing state files?

heroku/data_source_heroku_app.go
```diff
@@ -89,6 +91,32 @@ func dataSourceHerokuApp() *schema.Resource {
 					},
 				},
 			},
+
+			"team": {
+				Type:          schema.TypeMap,
```
bernerdschaefer: 👍

heroku/resource_heroku_app.go
```diff
@@ -321,13 +321,16 @@ func resourceHerokuTeamAppCreate(d *schema.ResourceData, meta interface{}) error
 
 	opts := heroku.TeamAppCreateOpts{}
 
-	if v, ok := d.GetOk("team"); ok {
-		v := d.Get("team").([]interface{})
-	} else if v, ok := d.GetOk("organization"); ok {
-		v := d.Get("organization").([]interface{})
+	var t []interface{}
+	if _, ok := d.GetOk("team"); ok {
+		t = d.Get("team").([]interface{})
```
bernerdschaefer: Should this be `d.Get("team").(map[string]interface{})`? It looks like `var tealDetails map[string]interface{}` instead of t could work for integrating the two models.

heroku/data_source_heroku_app.go
joestump: I'm not entirely sure. It's on my list to test though.

heroku/resource_heroku_app.go
joestump: One of these two are definitely off. Good catch!

joestump: Just took a second look. So this is `TypeList` rearing it's ugly head. The first thing is to get the _list_ of `team` or `organization` and then to extract the first entry as `map[string]interface`. It might be possible to tighten this up now that both are `TypeMap`.

talbright:
closing this due to inactivity: feel free to re-open if you wish to pursue this further.

PR #91, opened 2018-07-12
Add a heroku_space_inbound_ruleset resources for managing inbound CIDR blocks

joestump:
Introduces a new resource called `heroku_space_inbound_ruleset` that allows you to manage [inbound rulesets](https://devcenter.heroku.com/articles/platform-api-reference#inbound-ruleset-create) for a Heroku Private Space.

One of the oddities of this resource is that you can't delete a space's ruleset. It's like a singleton that defaults to `0.0.0.0/0`. I set it back to `0.0.0.0/0` when you "delete" the ruleset.

```hcl
provider "heroku" {}

resource "heroku_space" "foobar" {
  name         = "joe-testing-1"
  organization = "some-org-here"
  region       = "virginia"
}

resource "heroku_space_inbound_ruleset" "foobar" {
  space = "${heroku_space.foobar.name}"

  rule {
    action = "allow"
    source = "8.8.8.8/32"
  }

  rule {
    action = "allow"
    source = "8.8.8.0/24"
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.0.3 (Unreleased)
+## 1.1.0 (Unreleased)
 
 FEATURES:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 ## 1.0.3 (Unreleased)
+
+FEATURES:
+
+* r/heroku_space_inbound_ruleset: Add a new resource for managing [inbound IP rulesets](https://devcenter.heroku.com/articles/platform-api-reference#inbound-ruleset) for Heroku Private Spaces ([#91](https://github.com/terraform-providers/terraform-provider-heroku/pull/91))
+
 ## 1.0.2 (July 10, 2018)
 
 BUG FIXES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ FEATURES:
 
 * r/heroku_space_inbound_ruleset: Add a new resource for managing [inbound IP rulesets](https://devcenter.heroku.com/articles/platform-api-reference#inbound-ruleset) for Heroku Private Spaces ([#91](https://github.com/terraform-providers/terraform-provider-heroku/pull/91))
 
+IMPROVEMENTS:
+
+* HTTP headers are now shown when `TF_LOG` is set to `DEBUG` or `TRACE` [#89](https://github.com/terraform-providers/terraform-provider-heroku/pull/89)
+
 ## 1.0.2 (July 10, 2018)
 
 BUG FIXES:
--- a/heroku/resource_heroku_space_inbound_ruleset.go
+++ b/heroku/resource_heroku_space_inbound_ruleset.go
@@ -3,19 +3,10 @@ package heroku
 import (
 	"context"
 	"fmt"
-	"log"
-	"time"
 
 	heroku "github.com/cyberdelia/heroku-go/v3"
-	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
-	"github.com/hashicorp/terraform/helper/validation"
-	"github.com/terraform-providers/terraform-provider-aws/aws/validators"
 )
-
-type spacePeerInfo struct {
-	heroku.Peering
-}
 
 func resourceHerokuSpaceInboundRuleset() *schema.Resource {
 	return &schema.Resource{
@@ -40,11 +31,6 @@ func resourceHerokuSpaceInboundRuleset() *schema.Resource {
 						"action": {
 							Type:     schema.TypeString,
 							Required: true,
-							Default:  codebuild.CacheTypeNoCache,
-							ValidateFunc: validation.StringInSlice([]string{
-								"allow",
-								"deny",
-							}, false),
 						},
 						"source": {
 							Type:         schema.TypeString,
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
 ## 1.0.3 (Unreleased)
+
+FEATURES:
```
davidji99: Maybe we can bump to 1.1.0 since this is a new feature/resource?

heroku/resource_heroku_space_inbound_ruleset.go
```diff
@@ -0,0 +1,136 @@
...
+		}{
+			Action: data["action"].(string),
+			Source: data["source"].(string),
+		})
+	}
+
+	return heroku.InboundRulesetCreateOpts{Rules: ruleset}
+}
+
+func resourceHerokuSpaceInboundRulesetSet(d *schema.ResourceData, meta interface{}) error {
```
davidji99: Could we follow what seems the general convention for naming the create method? So `resourceHerokuSpaceInboundRulesetCreate` instead of `resourceHerokuSpaceInboundRulesetSet`?

Or was there a particular reason for naming it with `Set`?

heroku/resource_heroku_space_inbound_ruleset.go
```diff
@@ -0,0 +1,136 @@
...
+
+	rulesList := []interface{}{}
+	for _, rule := range ruleset.Rules {
+		values := map[string]interface{}{}
+		values["source"] = rule.Source
+		values["action"] = rule.Action
+		rulesList = append(rulesList, values)
+	}
+
+	d.SetId(ruleset.ID)
```
davidji99: Do we also want to store the `space` info? I see in https://devcenter.heroku.com/articles/platform-api-reference#inbound-ruleset that the API returns this:

```
  "space": {
    "id": "01234567-89ab-cdef-0123-456789abcdef",
    "name": "nasa"
  },
```

Maybe since you have `space` as an attribute, we can add `d.Set("space", Ruleset.Space.Name)`?

heroku/resource_heroku_space_inbound_ruleset.go
joestump: I named it `Set` because you don't ever _really_ "create" a ruleset; just manipulate one. Uniformity is a good idea though.

heroku/resource_heroku_space_inbound_ruleset.go
```diff
@@ -0,0 +1,136 @@
...
+		Action string `json:"action" url:"action,key"`
+		Source string `json:"source" url:"source,key"`
+	}{
+		Action: "allow",
+		Source: "0.0.0.0/0",
+	})
+
+	_, err := client.InboundRulesetCreate(context.TODO(), spaceIdentity, heroku.InboundRulesetCreateOpts{Rules: rules})
+	if err != nil {
+		return fmt.Errorf("Error resettting inbound ruleset for space (%s): %s", spaceIdentity, err)
```
davidji99: typo here `resettting`

heroku/resource_heroku_space_inbound_ruleset.go
```diff
@@ -0,0 +1,136 @@
...
+				Type:     schema.TypeSet,
+				Required: true,
+				MinItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"action": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"source": {
```
davidji99: I wonder if the `Type` can be `schema.TypeSet` here instead? Because then one could do

```
rule { 
    action = "allow"
    sources = ["8.8.8.8/32", "8.8.8.0/24"]
  }
```

instead of 
```
 rule {
   action = "allow"
    source = "8.8.8.0/24"
  }

 rule {
   action = "allow"
    source = "8.8.8.0/32"
  }
```

This would mean the the create/update method would have to iterate over the `sources`.

joestump: I did it this way so that people could do the Terraform hack of `rules = ["${var.list_of_rule_maps}"]` which magically works under the hood.

joestump: If we went that route, I'd propose we do this instead:

```hcl
resource "heroku_space_inbound_ruleset" "foobar" {
  space = "foobar"

  allow {
    sources = ["8.8.8.8/32", "8.8.8.0/24"]
  }

  deny {
    source = ["8.8.4.4/32"]
  }
}
```

joestump: Or just `allow = ["8.8.8.8/32", "8.8.8.0/24"]` and `deny = ["8.8.4.4/32"]`.

heroku/resource_heroku_space_inbound_ruleset.go
joestump: I remember now, I called it `Set` because it's used for both `Create` and `Update`. These two operations are identical from the API's perspective.

bernerdschaefer:
I don't have time to go deep, but the example TF in the PR description looks great and seems to map nicely back and forth from the schema definition. 👍

heroku/resource_heroku_space_inbound_ruleset.go
davidji99: okay works for me.

davidji99:
👍

PR #99, opened 2018-07-17
Migrate to dep for managing vendor/ and update packages in vendor/

joestump:
This moves `vendor/` management to Go's [`dep`](https://golang.github.io/dep/) tool. I started by removing the existing `vendor/` and then doing a `dep init` to recreate it along with a `Gopkg.toml` and `Gopkg.lock`. 

Very excited about this for package management. Here's `dep status`:

```
joestump-ltm3:terraform-provider-heroku joe.stump$ dep status
PROJECT                                    CONSTRAINT     VERSION        REVISION  LATEST    PKGS USED
github.com/agext/levenshtein               v1.2.1         v1.2.1         5f10fee   v1.2.1    1   
github.com/apparentlymart/go-cidr          branch master  branch master  2bd8b58   2bd8b58   1   
github.com/apparentlymart/go-textseg       branch master  branch master  b836f5c   b836f5c   1   
github.com/armon/go-radix                  branch master  branch master  1fca145   1fca145   1   
github.com/aws/aws-sdk-go                  v1.14.27       v1.14.27       94b8014   v1.14.27  31  
github.com/bgentry/go-netrc                branch master  branch master  9fd32a8   9fd32a8   1   
github.com/bgentry/speakeasy               v0.1.0         v0.1.0         4aabc24   v0.1.0    1   
github.com/blang/semver                    v3.5.1         v3.5.1         2ee8785   v3.5.1    1   
github.com/cyberdelia/heroku-go            branch master  branch master  8c8bc98   8c8bc98   1   
github.com/davecgh/go-spew                 v1.1.0         v1.1.0         346938d   v1.1.0    1   
github.com/fatih/color                     v1.7.0         v1.7.0         5b77d2a   v1.7.0    1   
github.com/go-ini/ini                      v1.38.1        v1.38.1        358ee76   v1.38.1   1   
github.com/golang/protobuf                 v1.1.0         v1.1.0         b4deda0   v1.1.0    5   
github.com/google/go-querystring           branch master  branch master  53e6ce1   53e6ce1   1   
github.com/hashicorp/errwrap               branch master  branch master  d6c0cd8   d6c0cd8   1   
github.com/hashicorp/go-cleanhttp          branch master  branch master  d5fe4b5   d5fe4b5   1   
github.com/hashicorp/go-getter             branch master  branch master  a33f09c   a33f09c   2   
github.com/hashicorp/go-hclog              branch master  branch master  ff2cf00   ff2cf00   1   
[...]

PR #100, opened 2018-07-18
Add `internal_routing` option to `heroku_app`

alindeman:
This pre-beta feature enables Heroku apps to be routable only within the private space. Specifically, the `foo.herokuapp.com` DNS name will resolve to a load balancer with an internal IP address, routable only within the private space.

While this feature is not generally available at this time, we ideally want the Terraform provider to support it at the time of launch.

I've also started transitioning the internals of `heroku_app` to use `Team`, rather than `Organization` since the term `organization` is deprecated in favor of `team`. They are functionally equivalent for most cases. However, new functionality (`internal_routing` included) is only being added to the `team` route, rather than the `organization` route to encourage the switchover. These changes shouldn't introduce any breaking changes for users of the provider though. Eventually, we might want to start moving towards the provider using the term `team` as well, but that is a larger change that must be planned carefully.

I've also removed the requirement for `HEROKU_EMAIL` to be set for acceptance tests because as long as `HEROKU_API_KEY` is set (which is already required), specifying an email is not necessary.

### DIFF
--- a/heroku/resource_heroku_app.go
+++ b/heroku/resource_heroku_app.go
@@ -16,12 +16,13 @@ import (
 
 // herokuApplication is a value type used to hold the details of an
 // application. We use this for common storage of values needed for the
-// heroku.App and heroku.OrganizationApp types
+// heroku.App and heroku.TeamApp types
 type herokuApplication struct {
 	Name             string
 	Region           string
 	Space            string
 	Stack            string
+	InternalRouting  bool
 	GitURL           string
 	WebURL           string
 	OrganizationName string
@@ -56,6 +57,9 @@ func (a *application) Update() error {
 		a.App.GitURL = app.GitURL
 		a.App.WebURL = app.WebURL
 		a.App.Acm = app.Acm
+		if app.InternalRouting != nil {
+			a.App.InternalRouting = *app.InternalRouting
+		}
 		if app.Space != nil {
 			a.App.Space = app.Space.Name
 		}
@@ -125,6 +129,13 @@ func resourceHerokuApp() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				Computed: true,
+			},
+
+			"internal_routing": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
 			},
 
 			"buildpacks": {
@@ -222,6 +233,9 @@ func resourceHerokuAppImport(d *schema.ResourceData, m interface{}) ([]*schema.R
 	// easy to get, so it makes more sense to just use the name as much as possible.
 	d.SetId(app.Name)
 	d.Set("acm", app.Acm)
+	if app.InternalRouting != nil {
+		d.Set("internal_routing", *app.InternalRouting)
+	}
 
 	return []*schema.ResourceData{d}, nil
 }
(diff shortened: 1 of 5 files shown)


heroku/resource_heroku_app.go
```diff
@@ -321,9 +335,14 @@ func resourceHerokuOrgAppCreate(d *schema.ResourceData, meta interface{}) error
 		log.Printf("[DEBUG] App stack: %s", vs)
 		opts.Stack = &vs
 	}
+	if v, ok := d.GetOk("internal_routing"); ok {
+		vs := v.(bool)
+		log.Printf("[DEBUG] App internal routing: %v", vs)
+		opts.InternalRouting = &vs
+	}
```
joestump: Looking at the docs, I'm thinking we should put a check here to ensure a `space` has been specified as well. If not, return an error.

website/docs/r/app.html.markdown
```diff
@@ -46,6 +46,9 @@ The following arguments are supported:
      configuration variables set externally won't be removed by Terraform
      if they aren't present in this list.
 * `space` - (Optional) The name of a private space to create the app in.
+* `internal_routing` - (Optional) If true, the application will be routable
```
davidji99: Since it's functionality isn't generally available, could we add a disclaimer that states this? Or such a disclaimer might not be needed if we merge/release this as close to this feature's GA release.

alindeman: I've add a disclaimer that we can remove when the feature is GA. What do you think?

joestump: @alindeman I'd move it up above the attributes into a `~> **NOTE:**` block. You can see an example [here](https://www.terraform.io/docs/providers/heroku/r/team_collaborator.html).

davidji99: Looks good. Thanks!

heroku/resource_heroku_app.go
davidji99: I think we could do this once #99 goes in and we can access `CustomizeDiff`. From Hashicorp:

```
As the name suggests its primary purpose is to _modify_ the diff (populating concrete values for computed attributes where possible, etc) but it can also return an error to block the creation of a plan
it involves writing a function of this `CustomizeDiffFunc` type, which can then access attributes using `d.Get` like you'd normally be able to do in e.g. the `Create`/`Update` functions
```

website/docs/r/app.html.markdown
alindeman: @joestump I'm not quite sure how to format this in a reasonable way if it's called out in a note above the other attributes. In the scheme of things, it's a relatively minor attribute. I'm not quite sure it makes sense to draw this much attention to it. However, if you have a specific thing in mind, let me know and I'll commit it.

heroku/resource_heroku_app.go
alindeman: Indeed, if an API request is made with `internal_routing: true` for a common spaces app (outside a private space), the API will return an error message and an HTTP status code 422.

website/docs/r/app.html.markdown
joestump: I'd put this above the attributes section:

```
~> **NOTE:** Internal routing is currently a beta feature and requires you email Heroku support enable it for your space. Without this feature enabled by Heroku Support the `internal_routing` attribute will result in the Heroku API will return errors.
```

joestump:
@davidji99 @alindeman `internal_routing` has been enabled on the `sfdc-ohana-terraform-sandbox` organization so we should be able to run acceptance tests for this attribute.

alindeman:
@joestump @davidji99 Thanks for the feedback so far. I've pushed some tweaks and a new acceptance test. Let me know what you're thinking.

davidji99:
Cool @alindeman, let me run the full build on this PR.

joestump:
@davidji99 if you're cool with this I say we :shipit: – Can you kick off an SFCI job to run the full test suite once merged to `master`?

PR #101, opened 2018-07-18
Ability to hide config vars from Plan Output

davidji99:
TODO:
- [x] Create
- [x] Update
- [x] Read
- [x] Delete
- [ ] Handle backwards compatibility with `heroku_app` config vars
- [x] Add validation to check if one config var is set in both arguments
- [x] Tests (especially to check for ouputs from `private` vars
- [ ] Document how to best handle moving from `heroku_app` config vars to new resource
- [ ] Mark `heroku_app` config vars as deprecated
- [ ]  Import

Example:

```hcl
resource "heroku_app_config_vars" "foobar" {
  app = "${heroku_app.foobar.name}"
   public {
    name = "RAILS_ENV"
    value = "staging"
  }
   private {
    name = "DATABASE_URL"
    value = "postgres://user:password@some-host.com:1234"
  }
}
```

### DIFF
--- a/heroku/helper.go
+++ b/heroku/helper.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/heroku/heroku-go/v3"
+	"reflect"
 )
 
 // getAppName extracts the app attribute generically from a Heroku resource.
@@ -64,4 +65,17 @@ func getAppUuid(appName string, client *heroku.Service) string {
 	app, _ := doesHerokuAppExist(appName, client)
 
 	return app.ID
+}
+
+func SliceExists(slice interface{}, item interface{}) bool {
+	s := reflect.ValueOf(slice)
+	if s.Kind() != reflect.Slice {
+		panic("SliceExists() given a non-slice type")
+	}
+	for i := 0; i < s.Len(); i++ {
+		if s.Index(i).Interface() == item {
+			return true
+		}
+	}
+	return false
 }
--- a/heroku/helper.go
+++ b/heroku/helper.go
@@ -41,6 +41,7 @@ func doesHerokuAppExist(appName string, client *heroku.Service) (*heroku.App, er
 		log.Println(err)
 		return nil, fmt.Errorf("[ERROR] Your app does not exist")
 	}
+
 	return app, nil
 }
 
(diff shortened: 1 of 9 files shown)


sbabu-salesforce:
Curious what the state of this PR is. I could help usher it through, as my team wants this change merged as well.

davidji99:
yeah @sbabu-salesforce, let me circle back to this this week.

davidji99:
I'm starting to wonder if we really need to make a distinction between a `public` or `private` type of app config var. The Heroku Platform API does not have types of config variables and I think generally, all config variables could be considered private/secret. However, I could see people needing to see the config_vars in their output though I wonder if that ability is outweighed by making sure actual secrets aren't outputted in plaintext in a CI build.

Plus, in building out this artificial distinction on the proposed standalone resource, I see a few complications:  

1. How would one import a `public` config var? I could potentially see the sample command as `terraform import heroku_app_config_var.foobar <app_id>:<public>:<config_vars_separated_by_commas>` or `terraform import heroku_app_config_var.foobar <app_id>:<private>:<config_vars_separated_by_commas>`. I'll test this out locally and see how feasible this import route would be for users.

2. How do we allow users to migrate from `heroku_app`'s `config_vars` to `heroku_app_config_var`. Having the user delete `config_vars` would be a non-starter since that would delete the actual variables from the app. The user could first import `heroku_app_config_var` and then somehow remove `config_vars` from their `heroku_app` resource in state. I'm not sure if this is at all possible with `terraform state rm <ITEM>`?

Thinking out loud here. Feel free to chime in.

davidji99:
@talbright @joestump @bernerdschaefer if you have any 2cents.

talbright:
I like the proposal. Strictly speaking no one would be able to straight-up import the sensitivity, given that the back-end doesn't support it. But that would be the case regardless of the HCL implementation.

General observation: from the various PRs I've seen, we should make a point of any changes to HCL have illustrative examples in the PR description. Maybe it's just me being a more visual person, but its super helpful.

davidji99:
> General observation: from the various PRs I've seen, we should make a point of any changes to HCL have illustrative examples in the PR description. Maybe it's just me being a more visual person, but its super helpful.

Will do @talbright.

talbright:
> However as your mentioned combined with my current implementation in this PR, this adds a bit too much complexity and comes with some usability issues. Perhaps adding the sensitive_config_vars attribute to heroku_app resource would be the simplest fix to addressing plaintexting config var secrets.

I concur -- keep this PR to addressing the sensitive var issue, and make splitting vars out a separate pr.

Issue #72, opened 2018-06-19
Ability to hide config vars from Plan Output


davidji99:
From @joslynesser,

> To give context on how we use this proposed change in our private fork:
> 
> Without a vault cluster (or some other centralized secret store), one can leverage heroku config vars and distribute those secrets to every app across all regions with terraform. This keeps a secret source to a single location and avoids human error.
> 
> One create sa heroku app that houses these secrets, and then fetch values using all_config_vars from that central app to distribute the secrets to all other apps. This because our backend terraform state only exists in memory and is encrypted at rest in S3. Secrets are fetched in memory on the terraform run from the state that houses the secrets app, and then set on all apps needing that secret value.
> 
> Therefore, we need to ensure our secrets are not plain texted shown in a plan/apply output on say a CI tool.

joslynessersfdc:
Thanks @davidji99!

This is definitely a highly subjective change, but it solved our immediate problem of having secrets shown in plan/apply output. Ideally we'd have something like `config_vars` vs `sensitive_config_vars` or some other notion of separating sensitive values from the set, as it makes reviewing plans more difficult. Currently we check the heroku app (or current state) vs the code diff to determine what the value is changing from/to, but ideally that would only be for changing secrets and not every single config var. Maybe related to how https://github.com/terraform-providers/terraform-provider-heroku/issues/47 pans out?

joslynesser:
Thanks @davidji99!

This is definitely a highly subjective change, but it solved our immediate problem of having secrets shown in plan/apply output. Ideally we'd have something like config_vars vs sensitive_config_vars or some other notion of separating sensitive values from the set, as it makes reviewing plans more difficult. Currently we check the heroku app (or current state) vs the code diff to determine what the value is changing from/to, but ideally that would only be for changing secrets and not every single config var. Maybe related to how #47 pans out?

davidji99:
One way would be this on `heroku_app`: 
![screen shot 2018-07-10 at 1 45 56 pm](https://user-images.githubusercontent.com/3991975/42536699-9e5339a6-8447-11e8-8d96-869d9525fa1e.png)

joslynesser:
I could see folks wanting to see non-sensitive config vars in the diff output. Maybe a separate `sensitive_config_vars` attribute to differentiate while still keeping the current design? Or a redesign like #47 that allows individual specification on sensitivity?

drn:
[...]

Issue #47, opened 2018-02-22
heroku_config resource


sheax0r:
Currently, the only way to manage config variables on a heroku app is via the `heroku_app` resource. ie:

```hcl
resource "heroku_app" "my_app" {
  name = "my-app"
  config_vars {
    CONFIG_VAR = "value"
  }
}
```

This approach has a few problems:

1. It results in centralization of config vars. All config vars coming in from any other generated resources get set *in one place*. This means that the setting of a config var on an app can be far away from the place the var is defined, which may make it harder to follow the flow.

2. It makes it impossible to set config vars on an app that isn't managed by terraform. This is somewhat dangerous, since it means that even if all you really want to do is set a few config vars on an app, you're obligated to give your script the ability to manage other properties of the app as well (and even to destroy or create it).

## Possible Solution: heroku_config resource

We add a `heroku_config` resource, like this:

```hcl
resource "heroku_config" "my_config_vars" {
  app = "my-app"
  config_vars {
    CONFIG_VAR = "value"
  }
}
```

You can create multiple resources to configure different sets of config vars on the same app. It also still supports management of many config vars in the same resource, which is good for keeping things minimally verbose (my initial idea was to have one resource per config var, but I feel like this is a bit better).

### Config var conflicts
What to do if config vars conflict across multiple resources? I suggest we detect this situation and fail if it happens.

### Release Explosion
What to do about possibly generating lots of releases and restarting apps repeatedly?
I suggest we defer the actual setting of config vars until the end of a run. Is this possible? If it's not possible, I suggest that we just document it - I don't think this is necessarily a blocker.

sheax0r:
@bernerdschaefer

joestump:
Regarding the first problem, you can always pass `config_vars = "${var.config_vars}"` rather than creating a block like that. We do this in our Heroku app TF module. 

I'd also like to raise the idea of setting a single variable with a single resource, which would allow us to us things like `count` for variable setting:

```hcl
resource "heroku_app_config_var" "GITHUB_TOKEN" {
  app = "${heroku_app.foo.name}"
  name = "GITHUB_TOKEN"
  value = "${var.GITHUB_TOKEN}"
}
```

bernerdschaefer:
The standalone config_var resource is an interesting one. For the default case, I think the current `config_vars` approach is the right one, but there are some cases where it breaks down.

[...]

PR #102, opened 2018-07-18
Better addon support

shirdoo:


heroku/resource_heroku_addon.go
```diff
@@ -154,6 +160,21 @@ func resourceHerokuAddonRead(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
+	appConfigInfo, err := client.ConfigVarInfoForApp(context.TODO(), addon.App.Name)
```
shirdoo: basically, pull down the configs from the parent app this addon is attached to, find the ones associated with this addon from the config_vars list, and add the key/value back to the addon resource

joestump:
@shirdoo thanks for the PR! Can you show me some example HCL of what this would enable that can't currently be done? It's not entirely clear what I'd do with all of the config vars from an app in the addon resource.

The only thing that springs to mind is passing service credentials/info into another app, but I was under the impression that's what the `heroku_addon_attachment` would do.

But I think I understand the main issue, which is that config vars that an add-on inserts into an application aren't available until the _next_ TF run.

joestump:
Okay, I think I get it now. `config_vars` is just a list of variables. This gets us a map of variables w/ values. I wonder why `config_vars` was just a list to begin with. @davidji99 @bernerdschaefer thoughts on making `config_vars` a map instead? It'd be a backwards compatibility break, but I feel like it's kind useless in list form.

shirdoo:
"thoughts on making config_vars a map instead?"

I want to say, this is the response from the heroku API, that's been ported directly into terraform resource schema: https://devcenter.heroku.com/articles/platform-api-reference#add-on-info

I can certainly convert config_vars to be a map, thats populated off the app's all_config_vars, just a matter of how close do we want the two schema's in sync

heroku/resource_heroku_addon.go
```diff
@@ -55,6 +55,11 @@ func resourceHerokuAddon() *schema.Resource {
 				},
 			},
 
+			"all_config_vars": {
+				Type:     schema.TypeMap,
+				Computed: true,
+			},
+
```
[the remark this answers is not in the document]
shirdoo: sure thing. Where does docs for resources live?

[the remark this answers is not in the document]
davidji99: under `/website` directory

davidji99:
Can you comment on why you couldn't use an `addon_attachment` here instead?

README.md
```diff
@@ -62,6 +62,7 @@ In order to run the full suite of Acceptance tests, run `make testacc`. You will
 * `HEROKU_ORGANIZATION` – The organization to run the tests against.
 * `HEROKU_SPACES_ORGANIZATION` – The organization to run the Heroku Private Space tests against. 
 * `HEROKU_API_KEY` – A valid API key that has access to the two organizations listed above.
+* `HEROKU_EMAIL` - The email associated with above API key
```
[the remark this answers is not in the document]
shirdoo: hrrmm, but I get test failures without setting HEROKU_EMAIL..

joestump:
@shirdoo thanks for that context. @davidji99 I think we should add this, but I think we should make a breaking change to `config_vars` so that it's a map instead of a list of variable names. If people still really need a list of variable names they can do `${keys(heroku_addon.foo.config_vars)}` to get it.

I feel like a list of variable names from an add-on is pretty useless. I'm struggling to think what people would have used that for to begin with inside of TF.

README.md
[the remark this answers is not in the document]
joestump: @shirdoo that's what @alindeman removed. He said we didn't need to have it set for tests. We can remove this from the docs here and defer to the related work in #100.

shirdoo:
@joestump @davidji99  anything I haven't addressed to merge this change in?

davidji99:
👀 ing

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ IMPROVEMENTS:
 
 * Update the `heroku-go` client to the latest version. [#97](https://github.com/terraform-providers/terraform-provider-heroku/pull/97)
 * Migrate to dep for managing `vendor/` and update packages in `vendor/` [#99](https://github.com/terraform-providers/terraform-provider-heroku/pull/99)
+* Better support for addon config referencing for downstream heroku apps. [#102](https://github.com/terraform-providers/terraform-provider-heroku/pull/102/files)
```
davidji99: Unfortunately, one downside of Hashicorp's release process is they look at the Changelog for versioning. `1.2.0` is already released so can you move this to underneath `1.3.0`? Leave the `(Unreleased)` as it is.

davidji99: Let me know when this is changed and I'll merge this in.

shirdoo: thanks @davidji99 , I think I followed that, take another look when you get a chance? Cheers!

davidji99:
Oh @shirdoo, sorry if my earlier comment was not clear, I meant this:

```
## 1.3.0 (Unreleased)

IMPROVEMENTS:

* Better support for addon config referencing for downstream heroku apps. [#101](https://github.com/terraform-providers/terraform-provider-heroku/pull/102/files)

## 1.2.0 (July 21, 2018)

IMPROVEMENTS:

* Update the `heroku-go` client to the latest version. [#97](https://github.com/terraform-providers/terraform-provider-heroku/pull/97)
* Migrate to dep for managing `vendor/` and update packages in `vendor/` [#99](https://github.com/terraform-providers/terraform-provider-heroku/pull/99)

FEATURES:
* Add `internal_routing` option to `heroku_app` [#100](https://github.com/terraform-providers/terraform-provider-heroku/pull/100)
```

So your improvement will be under `1.3.0` and the other stuff was already released as part of `1.2.0`

shirdoo:
drr, sorry, think I got it this time

website/docs/r/addon.html.markdown
```diff
@@ -52,7 +52,8 @@ The following attributes are exported:
 * `name` - The add-on name
 * `plan` - The plan name
 * `provider_id` - The ID of the plan provider
-* `config_vars` - The Configuration variables of the add-on
+* `config_vars` - List of config variable keys the add-on maintains and attaches to associated heroku apps
```
davidji99: Hey @shirdoo,

@joestump had a comment earlier:

> @shirdoo thanks for that context. @davidji99 I think we should add this, but I think we should make a breaking change to config_vars so that it's a map instead of a list of variable names. If people still really need a list of variable names they can do ${keys(heroku_addon.foo.config_vars)} to get it.
> I feel like a list of variable names from an add-on is pretty useless. I'm struggling to think what people would have used that for to begin with inside of TF.

(Adding it here since we can do a reply chain)

What do you think? Would you personally prefer two attributes: one for just keys and one for keys/values? Or do you prefer Joe's recommendation? I'm on the side of keeping both but since it's easy enough just get `keys`, maybe we condense to just one attribute under `config_vars`?

shirdoo: so I think its a good idea to deprecate config_vars, but I'm hesitant to reassign the type of it, since there will be remote state that will have a type mismatch, breaking backwards compatibility? would deprecating config_vars would need to be a major release type change?

davidji99: I think we can probably move forward with this PR as is and phase out `config_vars` later on. If we think of something better, we can add a PR before we cut a release with this PR in it.

The current `config_vars` does not sound useful at all, but I'd like to open an issue and get feedback on who actually uses this attribute in their `.tf` files. If anyone does, we'll need to drum up a plan to deprecate/remove/state migration the change. @joestump you cool with this?

`would deprecating config_vars would need to be a major release type change?` <-- no but if we actually removed it from the schema, we'd probably need to consider a major version bump, though personally a minor bump would probably suffice. It seems looking at the AWS provider's CHANGELOG briefly, minor bumps were used when attributes were removed from state.

joestump: The current `config_vars` is marked as `Computed: true` so I don't think it will cause any state file issues. State will update on the next `apply`. What _would_ break is if someone actually uses these, but a simple `${keys(heroku_addon.foo.config_vars)}` would fix the issue.

shirdoo: just got back from PTO, I can make the change to repurpose config_vars as a map, will test locally against our zero backend, and see how it blows up. 

Requiring clients to update their code to `${keys(heroku_addon.foo.config_vars)}` is a breaking change IMO, as that will silently break said clients if they do not tag their provider release. Might be fine if this provider isn't being used much, but as a client, I'd be bummed if minor releases break me like that.

shirdoo: for example, this terraform script would break: https://github.com/travis-infrastructure/terraform-stuff/blob/dcaacbe148c33c1be1ca61466ae11bc02e5c5966/production/heroku.tf, and this one would no longer be importable: https://github.com/devops-studentXX/lab02/blob/5d0326a0ddbb6dde779d22d4dfb6c5a8a0988514/terraform/heroku/production/terraform.tfstate

Sure its an easy fix, but annoying to force requiring a fix on a minor upgrade

shirdoo:
@davidji99  one more review when you get a chance please?

davidji99:
Hey @shirdoo, thanks for testing it out!

From your comments below:
screen shot 2018-08-07 at 8 54 51 pm

Since there are existing scripts, though not recent, it would be annoying for end-users to have to reconfigure their `tf` files for a minor version change. The right thing would be to major bump but I wonder if other providers have only minor bumped on attribute changes. So it seems the better route is to go back to having both attributes.

@bernerdschaefer What do you think? The original change was an additional attribute that stores the key,value for addon variables and the current change is to change an existing computed attribute altogether.

shirdoo:
@bernerdschaefer  the config_vars on this resource is as stable as config_vars on the heroku_resource itself. This is refreshing the addon resource with the config_vars it "owns" on the heroku_resource, where I would just refresh the state of the heroku_resource itself, but terraform doesn't like updating resources outside of the one under operation. 

After running `heroku addons:wait new resource`, I'd hope whatever config vars the heroku addon is adding is added, so I can actually use the addon. Equivalently, the terraform heroku provider runs the equivalent wait, by posting to the addon status, and attaches whatever config_vars are populated.

shirdoo:
@joestump  @bernerdschaefer  any objections to seeing this merge? I have addons in my teams registry that is blocking me to bring up a single environment with a single apply behind this PR. I'd rather not move the addons to a separate module, I'd still need to apply's or an apply and a refresh to reference config_vars attached by addons.

bernerdschaefer:
The changes themselves look good 👍 

I'm conflicted about the feature / behavior, though, for two reasons:

* I checked with the team that owns these endpoints and it looks like most addons are and _should_ be updating config vars before transitioning to `provisioned`, it is not _required_ for them to do so.
* I think we should minimize custom behaviors offered by resources that don't exist in the platform APIs. heroku_app is a big exception to this, but much of that is historical, and we've actually been breaking out "magical" attributes into their own resources. I'm not sure that working around an intentional limitation (un-shareable addons) is a good enough reason to customize this resource.

Pondering the second point has me thinking about the relation between this and #47 and #72. If we had an app config var _data_ resource, could this be modeled entirely in user terraform rather than implemented in the provider? Something like:

```hcl
resource "heroku_addon" "default" {
  app = "app1"
  plan = "heroku-redis:premium-1"
}

data "heroku_app_config_vars" "redis" {
  app = "${heroku_addon.default.app}"
  keys = ["${heroku_addon.default.config_vars}"] # optional, defaults to all keys
}

resource "heroku_app" "app2" {
  name = "app2"
  config_vars = {
    NORMAL = "var"
  }
  config_vars = "${data.heroku_app_config_vars.redis.config_vars}"
}
```

shirdoo:
I don't follow that example, `data.heroku_app_config_vars.redis.config_vars` is what type exactly? I'm getting `Error: data.heroku_app_config_vars.newrelic: Provider doesn't support data source: heroku_app_config_vars`

bernerdschaefer:
@shirdoo Sorry for the confusion! That data source does not _currently_ exist, but could be built on #47.

PR #104, opened 2018-07-20
Add `heroku_space_vpn_connection` to establish VPN connections

alindeman:
This beta feature enables IPSec-based VPN tunnels to be created between a network and a Heroku Private Space.

While this feature is not generally available at this time, we ideally want the Terraform provider to support it at the time of launch.

### DIFF
--- a/heroku/resource_heroku_space_vpn_connection.go
+++ b/heroku/resource_heroku_space_vpn_connection.go
@@ -1,0 +1,173 @@
+package heroku
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"net/url"
+	"time"
+
+	heroku "github.com/cyberdelia/heroku-go/v3"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceHerokuSpaceVPNConnection() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceHerokuSpaceVPNConnectionCreate,
+		Read:   resourceHerokuSpaceVPNConnectionRead,
+		Delete: resourceHerokuSpaceVPNConnectionDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"space": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"public_ip": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"routable_cidrs": {
+				Type:     schema.TypeSet,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Required: true,
+				ForceNew: true,
+			},
+
+			"space_cidr_block": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"ike_version": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+
+			"tunnels": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Optional: true,
+
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"ip": {
+							Type:     schema.TypeString,
+							Computed: true,
+							Optional: true,
+						},
+
+						"pre_shared_key": {
+							Type:     schema.TypeString,
+							Computed: true,
+							Optional: true,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func resourceHerokuSpaceVPNConnectionRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*heroku.Service)
+	space := d.Get("space").(string)
+
+	conn, err := client.VPNConnectionInfo(context.TODO(), space, d.Id())
+	if err != nil {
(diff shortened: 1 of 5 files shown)


amerine:
LGTM (from another Heroku employee)

alindeman:
The API issue has been fixed. This is ready for review:

```
go test -v ./heroku -run VPN -timeout 120m
=== RUN   TestAccHerokuVPNConnection_basic
--- PASS: TestAccHerokuVPNConnection_basic (1119.38s)
PASS
ok      github.com/terraform-providers/terraform-provider-heroku/heroku 1119.410s
```

heroku/resource_heroku_space_vpn_connection.go
```diff
@@ -0,0 +1,173 @@
...
+				if herr, ok := err.(*url.Error).Err.(heroku.Error); ok && herr.ID == "not_found" {
+					// VPN creation is eventually consistent
+					return nil, "pending", nil
+				}
+				return nil, "", fmt.Errorf("Error getting VPN status: %v", err)
+			}
+
+			return conn.ID, conn.Status, nil
+		},
+		Timeout: 45 * time.Minute,
```
davidji99: In several other resources, 20 minutes if often used. Could 20 minutes be used here or will this new resource take more than 20 min periodically?

heroku/resource_heroku_space_vpn_connection.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return fmt.Errorf("Error waiting for VPN to become available: %v", err)
+	}
+
+	d.SetId(buildCompositeID(space, id))
```
davidji99: Any particular reason to use `buildCompositeID()` over just `id := conn.ID`?  My guess is multiple VPN connection can share the same `UUID`?

heroku/resource_heroku_space_vpn_connection.go
```diff
@@ -0,0 +1,173 @@
...
+				ForceNew: true,
+			},
+
+			"public_ip": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"routable_cidrs": {
```
davidji99: Could we use `schema.TypeSet` here or does ordering matter for this attribute?

```
TypeSet implements set behavior and is used to represent an unordered collection of items, meaning that their ordering specified does not need to be consistent, and the ordering itself has no impact on the behavior of the resource.

The elements of a set can be any of the other types allowed by Terraform, including another schema. Set items cannot be repeated.
```

website/docs/r/space_vpn_connection.html.markdown
```diff
@@ -0,0 +1,43 @@
...
+# heroku\_space\_vpn\_connection
+
+Provides a resource for creating a VPN connection between a network and a Heroku Private Space. For more information, see [Private Spaces VPN Connection](https://devcenter.heroku.com/articles/private-space-vpn-connection?preview=1) in the Heroku DevCenter.
+
+~> **NOTE:** VPN connections are currently a beta feature that requires you email Heroku Support to have it enabled. Attempting to create a VPN connection without the feature enabled by Heroku Support will fail.
+
+## Example Usage
+
+```hcl
+resource "heroku_space_vpn_connection" "office" {
```
davidji99: Could you add an example HCL here for space?

```
resource "heroku_space" "default" {
  name         = "some_name"
  organization = "some_team"
  region       = "virginia"
}
```

alindeman:
@davidji99 Thanks for the review! I've pushed a new changeset with your recommendations. Let me know if you have more feedback.

heroku/resource_heroku_space_vpn_connection.go
davidji99: 👍

heroku/resource_heroku_space_vpn_connection.go
davidji99: Sounds good.

heroku/resource_heroku_space_vpn_connection.go
bernerdschaefer: @alindeman @davidji99 I think this resource needs to use a composite ID for `terraform import` to work, since we need both the space and ID attributes to construct the URL used to refresh state.

heroku/resource_heroku_space_vpn_connection.go
```diff
@@ -0,0 +1,173 @@
...
+
+	log.Printf("[DEBUG] Waiting for VPN (%s) to be allocated", d.Id())
+	stateConf := &resource.StateChangeConf{
+		Pending: []string{"pending", "provisioning"},
+		Target:  []string{"active"},
+		Refresh: func() (interface{}, string, error) {
+			conn, err := client.VPNConnectionInfo(context.TODO(), space, id)
+			if err != nil {
+				if herr, ok := err.(*url.Error).Err.(heroku.Error); ok && herr.ID == "not_found" {
+					// VPN creation is eventually consistent
```
bernerdschaefer: Did you run into this case while testing? As of the last few weeks, VPN connection info should be immediately available after creating it.

alindeman: @bernerdschaefer I did not run into it, no. Since I was using our internal provider as a blueprint, though, I brought this in. If it's no longer an issue, I'll back it out.

alindeman:
Test run with the latest code:

```
=== RUN   TestAccHerokuVPNConnection_basic
--- PASS: TestAccHerokuVPNConnection_basic (901.39s)
PASS
ok      github.com/terraform-providers/terraform-provider-heroku/heroku 901.422s
```

bernerdschaefer:
Thanks @alindeman!

PR #105, opened 2018-07-28
Various test improvements and fixes

talbright:
Various test improvements and cleanups and fixes to `heroku_space_app_access` import.


talbright:
@davidji99 I recommend we get this merged in and reviewed, and I can open up another PR for fixing imports, otherwise the tests on master aren't going to pass.

talbright:
This is ready for review

heroku/import_heroku_space_app_access_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccHerokuSpaceAppAccess_importBasic(t *testing.T) {
+	spaceName := fmt.Sprintf("tftest1-%s", acctest.RandString(10))
+	org := getTestSpaceOrganizationName()
+	testUser := getTestUser()
```
bernerdschaefer: I think it would be a bug to ever use `getTestSpaceOrganizationName` or `getTestUser` without also using the corresponding `testAccSkip` helpers as well. What do you think, then, about updating these getTest helpers to accept `t` and rolling the skip logic directly into them?

heroku/import_heroku_space_app_access_test.go
```diff
@@ -0,0 +1,36 @@
...
+		},
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckHerokuSpaceDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckHerokuSpaceAppAccessConfig_basic(spaceName, org, testUser, []string{"create_apps"}),
+			},
+			{
+				ResourceName:      "heroku_space_app_access.foobar",
+				ImportStateId:     fmt.Sprintf("%s:%s", spaceName, testUser),
```
bernerdschaefer: Did you consider using `buildCompositeID` here?

talbright: Nope...I didn't see that method! I will change to use it.

heroku/import_heroku_space_app_access_test.go
talbright: Actually when I was doing this it did smell like these two could be combined. I was a bit more shy about skipping a test outside of `PreCheck` as the pattern in this repo is to do all of that in `PreCheck` (https://www.terraform.io/docs/extend/testing/acceptance-tests/testcase.html#precheck) Any side affects we should be concerned about that you know of?

bernerdschaefer: I can't think of any reason not to skip right away if we know won't have the info we need to run the test, but maybe @joestump @davidji99 or @vanstee have thoughts?

davidji99: It seems that we could add the skip logic to L14 but seeing how the terraform docs say this:

```
PreCheck if non-nil, will be called before any test steps are executed. It is commonly used to verify that required values exist for testing, such as environment variables containing test keys that are used to configure the Provider or Resource under test.
```

my vote is to keep the skip logic in the `Precheck` for now.

heroku/resource_heroku_space_app_access.go
```diff
@@ -42,6 +42,15 @@ func resourceHerokuSpaceAppAccess() *schema.Resource {
 	}
 }
 
+//callback for schema.ResourceImporter
+func resourceHerokuSpaceAppAccessImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	space, email := parseCompositeID(d.Id())
+	d.Set("space", space)
```
davidji99: In addition to the existing attributes you already defined, I believe you're missing a `d.SetId` here.

talbright: The call here takes care of setting the attributes: https://github.com/terraform-providers/terraform-provider-heroku/blob/9a31050d377f66c4b2a80a8fce47444875533ce3/heroku/resource_heroku_space_app_access.go#L50

heroku/provider_test.go
```diff
@@ -60,20 +63,93 @@ func TestProviderConfigureUsesHeadersForClient(t *testing.T) {
...
+type TestConfig struct{}
+
+func NewTestConfig() *TestConfig {
+	return &TestConfig{}
 }
 
-func testAccSkipTestIfUserMissing(t *testing.T) {
-	if os.Getenv("HEROKU_TEST_USER") == "" {
-		t.Skip("HEROKU_TEST_USER is not set; skipping test.")
+func (t *TestConfig) GetWithAction(key TestConfigKey, testing *testing.T, notFoundAction TestConfigKeyNotFoundAction, defaultValue ...string) (val string) {
```
bernerdschaefer: I found the generalization and use of action enum here to make the config flow challenging to reason about.

I think with a helper function like this:

```go
// getEnv looks for each key in the os Env in order, returning the
// first non-zero val or ok = false.
getEnv(keys ...string) (val string, ok bool)
```

Then GetOrSkip and GetOrAbort could be implemented directly on that base, covering all the cases needed here with less indirection.

heroku/provider_test.go
```diff
@@ -60,20 +63,93 @@ func TestProviderConfigureUsesHeadersForClient(t *testing.T) {
...
+
+func (t *TestConfig) Get(key TestConfigKey, testing *testing.T, defaultValue ...string) (val string) {
+	return t.GetWithAction(key, testing, TestConfigKeyNotFoundNoopAction, defaultValue...)
+}
+
+func (t *TestConfig) GetOrSkip(key TestConfigKey, testing *testing.T, defaultValue ...string) (val string) {
+	return t.GetWithAction(key, testing, TestConfigKeyNotFoundSkipAction, defaultValue...)
+}
+
+func (t *TestConfig) GetOrAbort(key TestConfigKey, testing *testing.T, defaultValue ...string) (val string) {
```
bernerdschaefer: I'm used to `*testing.T` being the first argument to a test helper. Like context.Context, this makes it easier to scan and understand what special behaviors this function has.

heroku/provider_test.go
```diff
@@ -60,20 +63,93 @@ func TestProviderConfigureUsesHeadersForClient(t *testing.T) {
 	}
 }
 
-func testAccPreCheck(t *testing.T) {
-	if v := os.Getenv("HEROKU_API_KEY"); v == "" {
-		t.Fatal("HEROKU_API_KEY must be set for acceptance tests")
-	}
+type TestConfigKey int
```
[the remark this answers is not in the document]
talbright: This is a fairly solid use case for enums in general, and using an int is the idiomatic way to represent enums in Go. Also you get autocompletion, easy name refactoring, and its less error-prone overall.

talbright:
@bernerdschaefer incorporated your feedback, which seems to cleanup some things nicely.

heroku/provider_test.go
```diff
@@ -60,20 +63,101 @@ func TestProviderConfigureUsesHeadersForClient(t *testing.T) {
...
+	return
 }
 
-func testAccSkipTestIfUserMissing(t *testing.T) {
-	if os.Getenv("HEROKU_TEST_USER") == "" {
-		t.Skip("HEROKU_TEST_USER is not set; skipping test.")
+func (t *TestConfig) GetOrAbort(testing *testing.T, keys ...TestConfigKey) (val string) {
+	val = t.Get(keys...)
+	if val == "" {
+		testing.Fatal(fmt.Sprintf("stopping test: config %v must be set", keys))
```
bernerdschaefer: Nitpick: this could be reduced to `testing.Fatalf("stopping test: ....", keys)` (same for Skip->Skipf above)

heroku/provider_test.go
```diff
@@ -60,20 +63,101 @@ func TestProviderConfigureUsesHeadersForClient(t *testing.T) {
 	}
 }
 
-func testAccPreCheck(t *testing.T) {
-	if v := os.Getenv("HEROKU_API_KEY"); v == "" {
-		t.Fatal("HEROKU_API_KEY must be set for acceptance tests")
+type TestConfigKey int
+
+const (
+	TestConfigUserKey TestConfigKey = iota
```
bernerdschaefer: I'd still prefer these to just be constantized strings directly (`const TestConfigUserKey TestConfigKey = "HEROKU_TEST_USER"`), but it's not a blocker.

talbright:
@bernerdschaefer lets catchup on slack, there's some integration failure tests on master, and a couple more these code changes have seemed to introduced that I need to figure out.

talbright:
@bernerdschaefer + @davidji99 

The getters need to be outside of `PreCheck`. For some reason I couldn't chase down (maybe its a race condition) the config was being applied before the variables were being set in `PreCheck` causing test failures in some cases.

With that in mind its necessary to apply `TF_ACC` logic to the config getters, because we don't want or need the config when we aren't running acceptance tests (`make test`)

If we are good with the recent changes made, lets merge this in and if there are concerns or enhancements to testing lets address in another PR, as this one is getting long in the tooth and this value adds a decent amount of value as is.

PR #106, opened 2018-08-04
Update heroku dep to point at heroku repo

vanstee:
[cyberdelia/heroku-go](https://github.com/cyberdelia/heroku-go) has been moved to [heroku/heroku-go](https://github.com/heroku/heroku-go), so we should update our references here.

* Updated dep in Gopkg.toml (and ran `dep ensure`)
* Updated all package references in imports


alindeman:
Looks like this may have changed the alphabetization so `gofmt` needs to be run on the changed files.

vanstee:
@alindeman Thanks. Fixed!

talbright:
Good timing, I was just thinking about doing this as well. Question: what are everyones thoughts on pegging the version instead of leaving the revision at master?

alindeman:
> Good timing, I was just thinking about doing this as well. Question: what are everyones thoughts on pegging the version instead of leaving the revision at master?

Since the code is mostly generated, it doesn't strike me as particular valuable to peg it in this case. In fact, keeping up with `master` seems like the right thing to me, because we'll get alerted of any breaking changes in the Heroku API sooner by keeping up with master. Breaking changes in the API are [rare and announced early for production-stability resources](https://devcenter.heroku.com/articles/api-compatibility-policy#production) but they occasionally happen.

vanstee:
Yep I agree with keeping this pointed at `master`. Upgrading the client often is probably more important than any minor instability at this stage of the project.

alindeman:
> assuming the right thing happens will end up causing issues in the long run

Could you elaborate a bit more on what you're concerned about in this case? I think I would agree with you in scenarios where version information is available and the project is following SemVer or something like it. But heroku/heroku-go is an automatically generated client for the Heroku Platform API, and I think it's in our best interest to have the code reflect the current state of reality for that API.

talbright:
Say a developer wants to update a dependency, so they run:

`$ dep ensure -update github.com/heroku/heroku-go`

* No flexibility to update to a different version -- it always has to be head. What if a change to master is a breaking one here, but we want an earlier commit (`HEAD~1`) in the meantime?
* Casual updates to dependencies without mindful intent can break things. I run `-update` get master, think all as well and continue on my way. Meanwhile some side affect ensues.

alindeman:
> No flexibility to update to a different version -- it always has to be head. What if a change to master is a breaking one here, but we want an earlier commit (HEAD~1) in the meantime?

`dep` allows dependencies to be locked down to a specific SHA if needed.

> Casual updates to dependencies without mindful intent can break things. I run -update get master, think all as well and continue on my way. Meanwhile some side affect ensues.

In my experience, keeping up with the latest changes is a good default. If it breaks things, it's typically easier to fix it in small increments, rather than getting into a situation where a dependency is locked so far in the past that it's a huge effort to upgrade it (and its dependencies) because so much has changed.

I'm still not sure what we'd even lock to in this case. `heroku/heroku-go` is updated frequently as the Heroku API itself changes, and the Heroku API's behavior gets updated (albeit typically in backwards-compatible ways) regardless of whether we lock or not.

vanstee:
@talbright Are you cool with me merging this? Figured, since this PR doesn't change how we manage the heroku-go dep version, we could continue discussing on another issue or PR if you want to submit one with the change you're describing.

talbright:
Yeah...no worries.

PR #107, opened 2018-08-06
Replace custom validators with validators provided by terraform

vanstee:
* Replace pipeline stage validator with [string in slice validator](https://github.com/hashicorp/terraform/blob/master/helper/validation/validation.go#L73-L93)
* Replace cidr notation validator with [similar terraform provided validator](https://github.com/hashicorp/terraform/blob/master/helper/validation/validation.go#L150-L182)


vanstee:
Thanks!

PR #113, opened 2018-08-16
Support for ~/.netrc authentication

### LINKED ISSUE #95: Support for ~/.netrc authentication
I think it would be a nice addition to support `.netrc` authentication.

Here is a Heroku go client issue that concluded clients should handle netrc: https://github.com/bgentry/heroku-go/issues/4

There is a client library already included to handle netrc but it doesn't seem to be used in this provider: https://github.com/terraform-providers/terraform-provider-heroku/tree/7948ef837c3f00f0fe672fd64bb679e67cdd09e8/vendor/github.com/bgentry/go-netrc

Heroku docs on netrc: https://devcenter.heroku.com/articles/authentication#netrc-file-format

davidji99:
Resolves #95 

I shopped around to see if any major providers, namely AWS, implement netrc. That didn't seem to be the case. Luckily [an unofficial provider](https://github.com/Yelp/terraform-provider-signalform) implemented a few methods to do this. Instead of importing that entire lib to get those methods, I copied them over (subject to some minor tweaks) and credited them.

Providing a test for netrc proved a bit difficult with the way we were implementing `config.go` so I changed that file up which caused me to refactor all the files under `/heroku`.

One question I had remains around the ordering of authentication. My current implementation dictates this order to get the email/api_token:

1. `~/.netrc`
2. environment variables
3. explicitly passed variables aka if a `main.tf` had 
```
provider "heroku" {
  email   = "ops@company.com"
  api_key = "${var.heroku_api_key}"
}
``` 

I'd figure the last stop is if the user explicitly defines email/api_token.


davidji99:
I shopped around to see if any major providers, namely AWS, implement netrc. That didn't seem to be the case. Luckily [an unofficial provider](https://github.com/Yelp/terraform-provider-signalform) implemented a few methods to do this. Instead of importing that entire lib to get those methods, I copied them over (subject to some minor tweaks) and credited them.

Providing a test for netrc proved a bit difficult with the way we were implementing `config.go` so I changed that file up which caused me to refactor all the files under `/heroku`.

One question I had remains around the ordering of authentication. My current implementation dictates this order to get the email/api_token:

1. `~/.netrc`
2. environment variables
3. explicitly passed variables aka if a `main.tf` had 
```
provider "heroku" {
  email   = "ops@company.com"
  api_key = "${var.heroku_api_key}"
}
```

heroku/data_source_heroku_app.go
```diff
@@ -99,7 +98,7 @@ func dataSourceHerokuApp() *schema.Resource {
 }
 
 func dataSourceHerokuAppRead(d *schema.ResourceData, m interface{}) error {
-	client := m.(*heroku.Service)
+	client := m.(*Config)
```
talbright: If you change these all to `client := m.(*Config.Api)` it would require less code changes.

heroku/provider.go
```diff
@@ -76,14 +82,25 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
...
+		config.APIKey = apiKey.(string)
 	}
 
 	log.Println("[INFO] Initializing Heroku client")
-	return config.Client()
+	if err := config.loadAndInitialize(); err != nil {
+		return nil, err
+	}
+
+	return &config, nil
```
talbright: 👍 Doing this makes the provider configuration a lot more usable across the code base, vs before where we only returned the API client.

heroku/data_source_heroku_app.go
davidji99: Ah yeah, good point. However I think this way will give the code a little more flexibility if say something like your proposal for additional attributes on the provider were added later on.

davidji99:
I'm noticing an impossible scenario where I can't test a netrc provider test because I need to unset `HEROKU_EMAIL` & `HEROKU_API_KEY` but those two variables are required for the other tests. The test should still require the use of the env variable.

This PR is mostly ready to go but I'd like to add a netrc test but it doesn't seem possible.

talbright:
Would this work?

```go
email:=os.Getenv("HEROKU_EMAIL")
os.Unsetenv("HEROKU_EMAIL")
defer os.Setenv("HEROKU_EMAIL", email)
resource.Test(t, resource.TestCase{ //test stuff here })
```

talbright:
The other option would be to refactor `TestConfig` to load and store the config from the env in `NewTestConfig`. That way other tests can't "pollute" the config by monkeying with environment variables.

heroku/provider.go
```diff
@@ -94,3 +113,55 @@ func parseCompositeID(id string) (string, string) {
 	parts := strings.SplitN(id, ":", 2)
 	return parts[0], parts[1]
 }
+
+func readNetrcFile(config *Config) error {
```
joestump: There are at least two `.netrc` Golang libraries. I feel like we'd be better served using a proper library for this. One that perhaps handles path lookups? (e.g. `$HOME` and then `.`)

heroku/resource_heroku_addon.go
```diff
@@ -126,7 +126,7 @@ func resourceHerokuAddonCreate(d *schema.ResourceData, meta interface{}) error {
 }
 
 func resourceHerokuAddonRead(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*heroku.Service)
+	client := meta.(*Config)
```
joestump: It feels weird to call these variables `client` when they are not the actual client. My vote would be to do one of the following two:

* Rename all of these `client` variables to `config`.
* Bumping up against the limits of my Go skills, but if possible can we do `client := meta.(*Config).Api` here?

davidji99: Yeah you can:
```
Bumping up against the limits of my Go skills, but if possible can we do client := meta.(*Config).Api here?
```

Trent had a similar remark about this in an earlier comment above and I wrote:
```
Ah yeah, good point. However I think this way will give the code a little more flexibility if say something like your proposal for additional attributes on the provider were added later on.
```

This way, instead of having one resource do `client := meta.(*Config)` and anther resource do `client := meta.(*Config).Api`, they all do the former. So in subsequent usages of `client` later on, they access the `Api` or additional fields of the `Config` struct. What do yo think?

heroku/resource_heroku_team_member.go
```diff
@@ -57,7 +57,7 @@ func resourceHerokuTeamMemberImport(d *schema.ResourceData, meta interface{}) ([
 
 // Callback for schema Resource.Create and schema Resource.Update
 func resourceHerokuTeamMemberSet(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*heroku.Service)
+	client := meta.(*Config)
```
talbright: I know this is a little painful, but with a good editor it shouldn't be too bad. The name of this var should be `config` vs `client`. 🤓

heroku/resource_heroku_addon.go
talbright: Seems good to me, I prefer the name `config` though. The `client` is part of the config, so I find using that name misleading.

heroku/resource_heroku_team_member.go
davidji99: haha okay, i'll change this to `config := meta.(*Config)`

davidji99:
I tried a big rename from `client` to `config` but it doesn't _look_ right. Seems https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/config.go#L192-L204 uses something similar to what I originally had? @talbright @joestump

talbright:
This illustrates your point better I believe: https://github.com/terraform-providers/terraform-provider-aws/blob/12d26526ce293649978c98c5a178bf7f3d63f8ba/aws/resource_aws_acm_certificate.go#L80

In the AWS provider:

```go
acmconn := meta.(*AWSClient).acmconn
```

Which in this provider would translate to:

```go
client := meta.(*Config).API
_, err := client.AddOnInfo(context.TODO(), d.Id())
```

If that's what you are suggesting, it looks good to me. As a slight nit I would rename `API` to `Client`.

davidji99:
@talbright updated with your suggestion.

talbright:
That looks a lot better to me 👍

heroku/provider.go
```diff
@@ -78,14 +84,27 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
...
-	config := Config{
-		Email:   d.Get("email").(string),
-		APIKey:  d.Get("api_key").(string),
-		Headers: h,
+	config.Headers = h
+
+	err := readNetrcFile(&config)
+	if err != nil {
+		return nil, err
+	}
```
joestump: Does this mean if I don't have a `.netrc` file the provider will fail to work?

heroku/resource_heroku_addon_migrate.go
```diff
@@ -4,12 +4,11 @@ import (
 	"context"
 	"fmt"
 	"github.com/hashicorp/terraform/terraform"
-	"github.com/heroku/heroku-go/v3"
 	"log"
 )
 
 func resourceHerokuAddonMigrate(v int, is *terraform.InstanceState, meta interface{}) (*terraform.InstanceState, error) {
-	conn := meta.(*heroku.Service)
+	conn := meta.(*Config)
```
joestump: Should this be `meta.(*Config).Api`?

heroku/resource_heroku_app.go
```diff
@@ -644,7 +644,7 @@ func updateConfigVars(
 	return nil
 }
 
-func updateBuildpacks(id string, client *heroku.Service, v []interface{}) error {
+func updateBuildpacks(id string, config *heroku.Service, v []interface{}) error {
```
joestump: Should we revert this to `client` given we're using that name everywhere else?

heroku/resource_heroku_app.go
```diff
@@ -587,8 +587,8 @@ func retrieveBuildpacks(id string, client *heroku.Service) ([]string, error) {
 	return buildpacks, nil
 }
 
-func retrieveAcm(id string, client *heroku.Service) (bool, error) {
-	result, err := client.AppInfo(context.TODO(), id)
+func retrieveAcm(id string, client *Config) (bool, error) {
+	result, err := client.Api.AppInfo(context.TODO(), id)
```
joestump: For consistency's sake, should we consider using doing `client := meta.(*Config).Api` everywhere rather than having some be `client.Api`?

talbright: As far as the suggested precedence goes: 
> 1. ~/.netrc
> 2. environment variables
> 3. explicitly passed variables aka if a main.tf had

Shouldn't environment variables always have the highest precedence?

joestump: I think the HCL should take highest precedence. My vote would be:

1. Explicit HCL configuration in the provider.
2. Environment variables.
3. ~/.netrc

talbright: That works

davidji99: yeah the order of precedence right now is HCL -> env variables -> netrc as Joe mentiooned.

talbright: well played...

talbright:
@davidji99 is there anything I can do to help you get this across the finish line? I know you've had to make a few changes because of us pesky reviewers, and that can get frustrating 👿

davidji99:
ah @talbright going to get this done today and tomorrow. Sorry for the delay.

talbright:
> ah @talbright going to get this done today and tomorrow. Sorry for the delay.

No need to apologize! Just wanted to lend a hand if you needed it.

davidji99:
Okay, I think I updated everything to be consistent. Review again @talbright @joestump?

talbright:
Looking...

PR #119, opened 2018-08-21
Slug resource

mars:
This new resource implements ["deploying slugs"](https://devcenter.heroku.com/articles/platform-api-deploying-slugs), making it possible to launch apps directly from a Terraform config. (A **heroku_app_release** can now be created with a `slug_id` from this new **heroku_slug** resource.)

```hcl
resource "heroku_slug" "sushi" {
  app = "sushi"

  process_types = {
    web = "ruby server.rb"
  }

  file_path = "slug.tgz"
}
```

Optionally when `file_path` is set, a local [slug archive](https://devcenter.heroku.com/articles/platform-api-deploying-slugs#create-slug-archive) is uploaded & its `checksum` generated.

Supported & tested operations:

* **create**
* **read**
* **import**

Complete example:

```hcl
resource "heroku_app" "sushi" {
  name   = "sushi"
  region = "us"
}

resource "heroku_slug" "sushi" {
  app       = "${heroku_app.sushi.id}"
  file_path = "slug.tgz"

  process_types = {
    web = "ruby server.rb"
  }
}

resource "heroku_app_release" "sushi" {
  app     = "${heroku_app.sushi.id}"
  slug_id = "${heroku_slug.sushi.id}"
}

resource "heroku_formation" "sushi" {
  app        = "${heroku_app.sushi.id}"
  type       = "web"
  quantity   = 1
  size       = "Standard-1x"
  depends_on = ["heroku_app_release.sushi"]
}
```


davidji99:
Initial review: Don't forget to add the docs. They are located under `/website`.

mars:
📚 just pushed the docs @davidji99

mars:
I just realized that it might be nice to auto-generate the `checksum` (SHA256 hash, really) of the optional `file_path` file, when it's set.

website/docs/r/slug.html.markdown
```diff
@@ -0,0 +1,101 @@
...
+* `size` - Slug archive filesize in bytes
+* `stack_id` - [Heroku stack](https://devcenter.heroku.com/articles/stack) identifier
+* `stack_name` - [Heroku stack](https://devcenter.heroku.com/articles/stack) name
+
+## Import
+Existing slugs can be imported using the combination of the application name, a colon, and the slug ID.
+
+For example:
+```
+$ terraform import heroku_slug.foobar foobar
```
talbright: Nit: the example doesn't match the description (missing the `:` etc)

heroku/resource_heroku_slug.go
```diff
@@ -0,0 +1,269 @@
...
+		return fmt.Errorf("Error opening slug file_path: %s", err)
+	}
+	stat, err := file.Stat()
+	if err != nil {
+		return fmt.Errorf("Error stating slug file_path: %s", err)
+	}
+	defer file.Close()
+
+	httpClient := &http.Client{}
+	req, err := http.NewRequest(method, httpUrl, file)
```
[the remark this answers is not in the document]
mars: These slug S3 PUT requests don’t need a custom Transport. They’re working fine without transport customization.

 As far as I can tell, that transport config is specific to what Heroku Platform API needs.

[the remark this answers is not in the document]
mars: Yeah, that Heroku API transport do `req.Header.Set("Accept", "application/vnd.heroku+json; version=3")` which would certainly mess with the fragile S3 pre-signed PUT requests 😝

heroku/resource_heroku_slug.go
```diff
@@ -0,0 +1,269 @@
...
+		return fmt.Errorf("Error creating slug upload request: %s", err)
+	}
+	req.ContentLength = stat.Size()
+	log.Printf("[DEBUG] Upload slug request: %+v", req)
+	res, err := httpClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("Error uploading slug: %s", err)
+	}
+
+	b, err := httputil.DumpResponse(res, true)
```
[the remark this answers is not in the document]
mars: ✅

website/docs/r/slug.html.markdown
mars: ✅

mars:
Thank you @talbright & @davidji99 for reviews 😊

heroku/resource_heroku_slug.go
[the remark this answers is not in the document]
talbright: Very well, just a thought. Trying to make sure we are re-using code where we can.

talbright:
@mars @davidji99 any thoughts on this question:

> is there a use case for also providing a data_source?

mars:
@talbright re:

> is there a use case for also providing a data_source?

The only use-case I can imagine for a data_source is to **list an app's slugs**. Unfortunately, this is not supported by the Heroku Platform API.

So, I do not see the need for `data_source_heroku_slug` quite yet.

davidji99:
> The only use-case I can imagine for a data_source is to list an app's slugs. Unfortunately, this is not supported by the Heroku Platform API.

Our team would like that endpoint as well. But to address a data resource for this, I think it's fine without for now.

heroku/resource_heroku_slug.go
```diff
@@ -0,0 +1,307 @@
...
+			},
+
+			"commit_description": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"process_types": {
+				Type:     schema.TypeList,
```
davidji99: Does order matter for this resource? If not, could we use `TypeSet`?

website/docs/r/slug.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+---
+
+# heroku\_slug
+
+Provides a [Heroku Slug](https://devcenter.heroku.com/articles/platform-api-reference#slug)
+resource.
+
+This resource supports [Creating Slugs from Scratch](https://devcenter.heroku.com/articles/platform-api-deploying-slugs), making it possible to launch apps directly from a Terraform config.
+
+If `file_path` is not specified for a slug, then the slug archive must be uploaded to `blob.method` + `blob.url` by some other means. Otherwise, a release will not be possible because there's no executable code.
```
davidji99: Would it be better to have this attribute as required to reduce a step that needs to be done out of band prior to running terraform? Or do you foresee a need for someone to upload the slug then use `heroku_slug`?

If you want to keep this attribute as optional, could we move this disclaimer to under `Argument Reference`?

website/docs/r/slug.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+
+## Minimal Example
+
+All that is required to create a ready-to-release slug:
```
davidji99: One thing that might be helpful to add is if you use this resource, you need to build/create the slug first. This might be a helpful *Note*.

website/docs/r/slug.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+* `blob` - Slug archive (compressed tar of executable code)
+  * `method` - HTTP method to upload the archive
+  * `url` - Pre-signed, expiring URL to upload the archive
+* `buildpack_provided_description` - Description of language or app framework, `"Ruby/Rack"`
+* `checksum` - Hash of the slug for verifying its integrity, auto-generated when `file_path` is set to upload a slug archive, `SHA256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`
+* `commit` - Identification of the code with your version control system (eg: SHA of the git HEAD), `"60883d9e8947a57e04dc9124f25df004866a2051"`
+* `commit_description` - Description of the provided commit
+* `process_types` - Map of [processes to launch on Heroku Dynos](https://devcenter.heroku.com/articles/process-model)
+* `size` - Slug archive filesize in bytes
+* `stack_id` - [Heroku stack](https://devcenter.heroku.com/articles/stack) identifier
```
davidji99: I'm a bit confused as to why one would need to export `stack_id` and `stack_name` when you also have `stack` as as an argument attribute. Do we need all three?

heroku/resource_heroku_slug.go
```diff
@@ -0,0 +1,307 @@
...
+	d.Set("app", app)
+	setState(d, slug)
+
+	return []*schema.ResourceData{d}, nil
+}
+
+func resourceHerokuSlugCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*heroku.Service)
+
+	app := d.Get("app").(string)
```
davidji99: Can use https://github.com/terraform-providers/terraform-provider-heroku/blob/master/heroku/helper.go#L12 here.

heroku/resource_heroku_slug.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+	defer file.Close()
+
+	httpClient := &http.Client{}
+	req, err := http.NewRequest(method, httpUrl, file)
+	if err != nil {
+		return fmt.Errorf("Error creating slug upload request: %s", err)
+	}
+	req.ContentLength = stat.Size()
+	if os.Getenv("TF_LOG") == "TRACE" || os.Getenv("TF_LOG") == "DEBUG" {
```
davidji99: I don't think you need to set this conditional. I believe by default, one doesn't see these logs until you do `TF_LOG=DEBUG` without this conditional.

website/docs/r/slug.html.markdown
mars: ✅ revised per suggestion

I think it's beneficial to keep `file_path` optional, as it leaves the door open to workflows where an external build system performs the slug upload.

That said, I don't think I would ever want to do that, but the last thing I want to do is limit utility of this new slug resource to only my own pre-conceived notions of how it will be used.

website/docs/r/slug.html.markdown
mars: ✅ added *Note* per suggestion

website/docs/r/slug.html.markdown
mars: This is a weird, legacy Heroku API artifact:
* when creating a Slug, the `stack` name argument is an optional string
* when reading a Slug, the `stack` attribute is a computed map of `stack_id` & `stack_name` where `stack_name` contains the original/create `stack` argument value

Is there a better way to express (or hide) this read vs write inconsistency in the resource?

I don't feel like hiding the `stack_id` would help anyone, but might be a problematically missing element is some unforeseen situation. Happy to adjust further based on feedback!

heroku/resource_heroku_slug.go
mars: ✅ now using `getAppName()`

heroku/resource_heroku_slug.go
mars: ✅ I reverted that `TF_LOG` stuff which I added based on feedback earlier in this PR

heroku/resource_heroku_slug.go
```diff
@@ -0,0 +1,307 @@
...
+	d.Set("checksum", slug.Checksum)
+	d.Set("commit", slug.Commit)
+	d.Set("commit_description", slug.CommitDescription)
+	processTypes := []map[string]string{slug.ProcessTypes}
+	if err := d.Set("process_types", processTypes); err != nil {
+		log.Printf("[WARN] Error setting process_types: %s", err)
+	}
+	d.Set("size", slug.Size)
+	d.Set("stack_id", slug.Stack.ID)
+	d.Set("stack_name", slug.Stack.Name)
```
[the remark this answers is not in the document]
mars: Stack names do not change. (See my prior comment about `stack`/`stack_name`/`stack_id.)

heroku/resource_heroku_slug.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+	return checksum, nil
+}
+
+func setState(d *schema.ResourceData, slug *heroku.Slug) error {
+	blob := []map[string]string{{
+		"method": slug.Blob.Method,
+		"url":    slug.Blob.URL,
+	}}
+	if err := d.Set("blob", blob); err != nil {
```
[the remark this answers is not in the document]
mars: Mostly for debugging as I'm learning about how Go allows runtime type errors 😣

(Why doesn't the compiler catch this stuff? Kind of befuddled why Go so eagerly allows type errors and pointer segfaults at runtime.)

mars:
I think this is ready for merge 🏁 unless further feedback requires changes.

heroku/resource_heroku_slug.go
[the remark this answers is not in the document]
talbright: Not that we are doing a good job of it in this code base, but you can use type assertions to help minimize type errors.

mars:
>> The only use-case I can imagine for a data_source is to list an app's slugs. Unfortunately, this is not supported by the Heroku Platform API.
>
> Our team would like that endpoint as well. But to address a data resource for this, I think it's fine without for now.

Slug IDs are included in App Releases. Would a new `data_source_heroku_app_release` be a viable solution to get previously released slugs? Or do you need/want to list all of an app's slugs released & unreleased?

davidji99:
👀 again

mars:
I'm looking into one final minor addition, to make it automatically force a new resource if the content of slug archive at `file_path` changes.

Please don't merge yet 🙏😆

website/docs/r/slug.html.markdown
davidji99: Hey @mars,

For `when creating a Slug, the stack name argument is an optional string`, can the value be the stack _id or stack_name? If this is possible, I am wondering if we can simplify this down to use justing the stack_id.

heroku/resource_heroku_slug.go
```diff
@@ -0,0 +1,328 @@
...
+			},
+
+			"size": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+
+			// Create argument; equivalent value as `stack_name` or `stack_id`
+			// depending on whether a UUID or stack name is set.
+			"stack": {
```
davidji99: I still wonder if we could just use `stack_id` instead of having all three.

mars: As I mentioned earlier:

> I wish this could be simplified like you suggest, but it would only serve to mislabel these values.
>
> When creating a slug `stack` accepts a **UUID** or **human-friendly name**.
>
> When reading a slug those two values are split into `stack_id` & `stack_name`.
>
> If we try to collapse any of them into each other, then their names do not reflect what they are anymore.

I'm frankly not sure when the stack UUID would be useful. Only trying to make this resource as useful as the API itself.

If this is a blocker to have this merged, I would be okay with reducing to only `stack_name` as the human-readably stack names are the interface we present in docs/examples.

mars: There are quite a few places in Heroku Platform API that have this interchangeable **name** or **ID** behavior. App name & UUID interchangeability is a big one.

Looking through the provider's existing resources, I see [pipeline_coupling](https://www.terraform.io/docs/providers/heroku/r/pipeline_coupling.html):

* accepts `app` argument
* returns `app` & `app_id` attributes.

Is this a good pattern to follow?

mars: What I'm suggesting is to simply remove `stack_name`, so this Slug provider would:

* accepts `stack` argument
* returns `stack` & `stack_id` attributes.

mars: I've updated this PR based on my [previous comment](https://github.com/terraform-providers/terraform-provider-heroku/pull/119#discussion_r215390088).

PR #120, opened 2018-08-29
Case insensitive logging check for TF_LOG

talbright:


talbright:
@bernerdschaefer this will fix the issue you had the other day, sorry for the frustrations that may have caused. The other fix in heroku/heroku-go#34 is still relevant -- this fixes an issue where in some cases the body output of an HTTP response is not visible.

heroku/config.go
```diff
@@ -17,7 +18,7 @@ type Config struct {
 // Client returns a new Service for accessing Heroku.
 func (c *Config) Client() (*heroku.Service, error) {
 	var debugHTTP = false
-	if os.Getenv("TF_LOG") == "TRACE" || os.Getenv("TF_LOG") == "DEBUG" {
+	if strings.EqualFold(os.Getenv("TF_LOG"), "trace") || strings.EqualFold(os.Getenv("TF_LOG"), "debug") {
```
bernerdschaefer: It looks like there's a [IsDebugOrHigher](https://godoc.org/github.com/hashicorp/terraform/helper/logging#IsDebugOrHigher) helper that could work here, and handles case-sensitivity.

talbright: Oh nice catch.

Issue #71, opened 2018-06-13
Add Exponential Backoff When adding Heroku Domain


davidji99:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
- v0.11.7

### Affected Resource(s)
- heroku_domain

### Terraform Configuration Files
```hcl
resource "heroku_app" "app" {
  name = "meh"

  region = "validregion"
  space  = "validspace"
}

resource "heroku_domain" "app" {
  app      = "${heroku_app.app.name}"
  hostname = "validfqdn"
}

```

### Output
heroku_domain.app: 1 error(s) occurred:
* heroku_domain.app: Post https://api.heroku.com/apps/meh/domains: App meh is being provisioned. Please try again later.

### Expected Behavior
It should wait until the app completes provisioning before trying to apply the domain change

### Actual Behavior
Failed apply

davidji99:
Proposed solution will be add some exponential backoff potentially.

talbright:
I ran into a form of this the other week as well. In my case the app already existed in the statefile, but I was adding `heroku_domain`. Of course re-running apply can get one past the problem, but that's far from ideal and not expected. 

The aws provider client has some built-in retry/backoff mechanism, we might look to see if that can be extracted. I've also use https://github.com/cenkalti/backoff before, but I would like to get some additional 👀 on it to see if that looks like a dependency we want to bring in. 

Ideally, I'd like to see retry logic built into the client, that way we get this for free across all our API calls and don't have to retrofit retry logic everywhere. I'll open a discussion in https://github.com/heroku/heroku-go.

talbright:
Actually perhaps the right thing to do here is use this: https://godoc.org/github.com/hashicorp/terraform/helper/resource#StateRefreshFunc

bernerdschaefer:
I think using StateRefreshFunc is the more appropriate solution for this case, rather than baking retries into the client.

talbright:
Agreed, the more I think about it the more that seems like the right thing to do.

bernerdschaefer:
[...]

PR #123, opened 2018-09-11
Retry space delete when apps still exist

ryanbrainard:
When deleting apps in a space and then deleting the space itself very rapidly, like `terraform destroy` would do, it is common that the apps aren't completely deleted when the space delete is attempted. Retrying a few seconds later usually resolves the issue, so this adds a retry for one minute. 

Please provide any guidance on the best way to test this. I looked at the other uses of `resource.Retry`, but it doesn't look like any of them are tested expictly.

### DIFF
--- a/heroku/resource_heroku_space.go
+++ b/heroku/resource_heroku_space.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	heroku "github.com/heroku/heroku-go/v3"
+	"strings"
 )
 
 type spaceWithRanges struct {
@@ -212,9 +213,24 @@ func resourceHerokuSpaceDelete(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*heroku.Service)
 
 	log.Printf("[INFO] Deleting space: %s", d.Id())
-	_, err := client.SpaceDelete(context.TODO(), d.Id())
-	if err != nil {
-		return err
+
+	// When deleting apps in a space and then the space itself,
+	// it is common that the apps aren't completely deleted
+	// when the space delete is attempted. Retrying a few seconds later
+	// usually resolves the issue, so retry for a minute.
+	retryError := resource.Retry(time.Minute, func() *resource.RetryError {
+		if _, err := client.SpaceDelete(context.TODO(), d.Id()); err != nil {
+			if strings.Contains(err.Error(), "Apps exist in this space") {
+				return resource.RetryableError(err)
+			}
+			return resource.NonRetryableError(err)
+		}
+
+		return nil
+	})
+
+	if retryError != nil {
+		return retryError
 	}
 
 	d.SetId("")


heroku/resource_heroku_space.go
```diff
@@ -212,9 +213,24 @@ func resourceHerokuSpaceDelete(d *schema.ResourceData, meta interface{}) error {
...
 	log.Printf("[INFO] Deleting space: %s", d.Id())
-	_, err := client.SpaceDelete(context.TODO(), d.Id())
-	if err != nil {
-		return err
+
+	// When deleting apps in a space and then the space itself,
+	// it is common that the apps aren't completely deleted
+	// when the space delete is attempted. Retrying a few seconds later
+	// usually resolves the issue, so retry for a minute.
+	retryError := resource.Retry(time.Minute, func() *resource.RetryError {
```
talbright: @davidji99 @bernerdschaefer do you think this is a better way to handle #71 (by checking for "app not ready" in the dns resource) over providing an arbitrary sleep in `heroku_app`?

talbright:
IMO there isn't a good way to test this, as long as integration tests pass that should be good enough.

mars:
Thank you @ryanbrainard! I witnessed this not-deleted Private Space recently and due to circumstances (I’d cleared the terminal history) was simply baffled!

heroku/resource_heroku_space.go
davidji99: I am wondering if it would be less 'resource creep' or isolated if we modified `heroku_app`'s DELETE function to check if the app was truly deleted or not. Maybe we can `GET` the `heroku_app` right after we `DELETE` it and wait exponentially for a `404` response. But that's assuming the heroku_app's `GET` call returns a 2XX right after the `DELETE`.

I suggest this instead since a child resource of space, the app, is the culprit here. I would like to stay away from arbitrary sleeps unless said sleeps were customizable which should be once I can get my netrc PR in which changes `config.go`.

davidji99:
This PR and a few other issues all point to a common issue of resources still available or not available after a DELETE or POST respectively. I think we'll need to come up with a common way of checking when the API does not provide any response indicating status. I think this check should be done exponentially to save on API calls with a maxTimeout that is user defined.

heroku/resource_heroku_space.go
ryanbrainard: > Maybe we can GET the heroku_app right after we DELETE it and wait exponentially for a 404 response. But that's assuming the heroku_app's GET call returns a 2XX right after the DELETE. What do you guys think?

I considered doing something like that; however, I think there might (just a theory) be something going on with caching and/or eventual consistency where not all API servers are immediately aware of the change. If this is actually true, what I'd be afraid of is that we'd get lucky and get a 404, but then hit a different server for the space delete and still get the error. Because of this, I went with handling the error when we see it.

talbright:
Unfortunately I forgot about this PR while working on #142. I'll drop another PR to add a `post_space_delete_delay` schema element in the `delays` HCL block.

talbright:
closing this due to inactivity: feel free to re-open if you wish to pursue this further.

PR #127, opened 2018-10-04
Export the UUID of the Heroku app.

joestump:
A use case was brought up in the `#terraform` channel today where someone was trying to use `null_resource` with `heroku_app.foo.id`, which is the app's name. The problem is that the `null_resource` triggers don't change even when the app is torn down and recreated.

This PR adds the UUID of the app to the exported attributes. Using this instead of `name` will allow `null_resource` triggers to fire when an app is recycled by TF.

### DIFF
--- a/heroku/resource_heroku_app_test.go
+++ b/heroku/resource_heroku_app_test.go
@@ -28,6 +28,8 @@ func TestAccHerokuApp_Basic(t *testing.T) {
 					testAccCheckHerokuAppAttributes(&app, appName, "heroku-16"),
 					resource.TestCheckResourceAttr(
 						"heroku_app.foobar", "name", appName),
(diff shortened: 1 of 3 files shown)


heroku/resource_heroku_app_test.go
```diff
@@ -28,6 +28,8 @@ func TestAccHerokuApp_Basic(t *testing.T) {
 					testAccCheckHerokuAppAttributes(&app, appName, "heroku-16"),
 					resource.TestCheckResourceAttr(
 						"heroku_app.foobar", "name", appName),
+					resource.TestCheckResourceAttrSet(
+						"heroku_app.foobar", "uuid"),
```
joestump: @shirdoo I check that `uuid` is set here.

PR #133, opened 2018-10-23
Set private_key parameter to sensitive

davidji99:


davidji99:
Per the code comment:

> 	// Sensitive ensures that the attribute's value does not get displayed in
	// logs or regular output. It should be used for passwords or other
	// secret fields. Future versions of Terraform may encrypt these
	// values.

I think this field should be set to `Sensitive`. Do you agree @joestump @talbright ?

joslynesser:
👍

PR #136, opened 2018-10-30
Fix formation.html.markdown

masutaka:
They seems required arguments.

https://github.com/terraform-providers/terraform-provider-heroku/blob/v1.5.0/heroku/resource_heroku_formation.go#L47-L61


davidji99:
thanks @masutaka

PR #137, opened 2018-10-31
Investigate & Fix the Following Test Failures

davidji99:
The following three test have been failing for a bit:

```
03:30:49 --- FAIL: TestAccHerokuSpaceAppAccess_importBasic (430.71s)
03:30:49 	testing.go:518: Step 0 error: Error applying: 1 error occurred:
03:30:49 			* heroku_space_app_access.foobar: 1 error occurred:
03:30:49 			* heroku_space_app_access.foobar: Patch https://api.heroku.com/spaces/tftest1-keuns4nk9t/members/****: Couldn't find that user.
```

```
03:30:49 --- FAIL: TestAccHerokuDomain_Basic (1.94s)
03:30:49 	testing.go:518: Step 0 error: Check failed: Check 5/5 error: heroku_domain.foobar: Attribute 'cname' expected "terraform-tftest-jxbeyywub0.example.com.herokudns.com", got "mechanistic-gorilla-tgcqkphfyw032xebcmam2ng3.herokudns.com"
```

```
03:30:49 --- FAIL: TestAccHerokuSpaceAppAccess_Basic (440.84s)
03:30:49 	testing.go:518: Step 0 error: Error applying: 1 error occurred:
03:30:49 			* heroku_space_app_access.foobar: 1 error occurred:
03:30:49 			* heroku_space_app_access.foobar: Patch https://api.heroku.com/spaces/tftest1-2ihxw8oxg6/members/****: Couldn't find that user.
03:30:49 		
03:30:49 		
03:30:49 		
03:30:49 	
```

We need to investigate and figure if the cause of these failures.

### DIFF
--- a/heroku/resource_heroku_domain_test.go
+++ b/heroku/resource_heroku_domain_test.go
@@ -31,8 +31,6 @@ func TestAccHerokuDomain_Basic(t *testing.T) {
 						"heroku_domain.foobar", "hostname", "terraform-tftest-"+randString+".example.com"),
 					resource.TestCheckResourceAttr(
 						"heroku_domain.foobar", "app", appName),
-					resource.TestCheckResourceAttr(
-						"heroku_domain.foobar", "cname", "terraform-tftest-"+randString+".example.com.herokudns.com"),
 				),
 			},
 		},
@@ -61,6 +59,10 @@ func testAccCheckHerokuDomainAttributes(Domain *heroku.Domain) resource.TestChec
 	return func(s *terraform.State) error {
 		if !strings.HasPrefix(Domain.Hostname, "terraform-") && !strings.HasSuffix(Domain.Hostname, ".example.com") {
 			return fmt.Errorf("Bad hostname: %s", Domain.Hostname)
+		}
+
+		if !strings.Contains(*Domain.CName, ".herokudns.com") {
+			return fmt.Errorf("Expected hostname to be [*.herokudns.com] but got: [%s]", *Domain.CName)
 		}
 
 		return nil


heroku/resource_heroku_domain_test.go
```diff
@@ -31,8 +31,6 @@ func TestAccHerokuDomain_Basic(t *testing.T) {
 						"heroku_domain.foobar", "hostname", "terraform-tftest-"+randString+".example.com"),
 					resource.TestCheckResourceAttr(
 						"heroku_domain.foobar", "app", appName),
-					resource.TestCheckResourceAttr(
-						"heroku_domain.foobar", "cname", "terraform-tftest-"+randString+".example.com.herokudns.com"),
```
davidji99: @talbright & @joestump - Curious how this test was working before. It doesn't seem we are able to determine what the CNAME is set to beforehand? Looking at the last several overnight test runs, the computed `cname` is always some random characters. I'd like to remove this line in the test for now unless I'm missing something else here?

davidji99:
It seems that we use the same user for `TestConfigNonAdminUserKey` & `TestConfigUserKey, which isn't a bad thing. However we have a team_member test that removes the same user that is needed to for the space_access. We might need to add test specific variables for users.

heroku/resource_heroku_domain_test.go
bernerdschaefer: Ah, yes, this changed recently: https://devcenter.heroku.com/changelog-items/1488

davidji99: ah gotcha. good to know. thanks @bernerdschaefer 

Wonder how we can leave this line in the test without knowing the `CNAME` beforehand.

davidji99: I guess what you suggested would do for now.

PR #139, opened 2018-11-02
Fetch slug archives via HTTPS

mars:
We've been working on the **Launch a Heroku app via Terraform** story. Adding the `heroku_slug` resource got us 90% of the way there. It currently stops short by requiring that the slug archive already be located on the local filesystem.

This PR adds a `heroku_slug.file_url` attribute, to configure the HTTP URL of the slug archive. When the resource is created, the specified URL is downloaded to a unique, temporary file, and then treated as if it was set in `file_path`.

```hcl
resource "heroku_slug" "foobar" {
  app                            = "foobar"
  file_url                       = "https://marsikai.s3.amazonaws.com/slugs/ruby-hello-world.tgz"
  buildpack_provided_description = "Ruby"

  process_types = {
    web = "ruby server.rb"
  }
}
```

Based on discussion in this PR, the `heroku_slug` resource now requires either `file_path` or `file_url`, removing the complexity-inducing, upload-it-yourself option.

### DIFF
--- a/heroku/resource_heroku_slug.go
+++ b/heroku/resource_heroku_slug.go
@@ -43,9 +43,10 @@ func resourceHerokuSlug() *schema.Resource {
 
 			// https:// URL of tarball to upload into slug
 			"file_url": {
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
+				Type:          schema.TypeString,
+				ConflictsWith: []string{"file_path"},
+				Optional:      true,
+				ForceNew:      true,
 			},
 
 			"blob": {
@@ -162,13 +163,14 @@ func resourceHerokuSlugCreate(d *schema.ResourceData, meta interface{}) error {
 		opts.BuildpackProvidedDescription = heroku.String(v.(string))
 	}
 
-	// Optionally, download the archive from an HTTP server
-	var downloadFilePath string
-	var userFilePath string
+	// Only file_path or file_url will be set, because of ConflictsWith.
 	var filePath string
+	// Simply use the configured file path
 	if v, ok := d.GetOk("file_path"); ok {
-		userFilePath = v.(string)
+		filePath = v.(string)
 	}
+	// Download the slug archive to a unique file path and clean it up
+	// after uploading to Heroku platform.
 	if v, ok := d.GetOk("file_url"); ok {
 		fileUrl := v.(string)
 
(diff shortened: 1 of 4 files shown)


heroku/resource_heroku_slug.go
```diff
@@ -40,6 +41,13 @@ func resourceHerokuSlug() *schema.Resource {
 				ForceNew: true,
 			},
 
+			// https:// URL of tarball to upload into slug
+			"file_url": {
```
davidji99: Do you want to allow users to set both `file_url` and `file_path`? It seems one or the other not both? If so, I'd suggest you do a `ConflictsWith: []string{"file_path"}` and vice versa?

Do you also want to use `DiffSuppressFunc` to not trigger `ForceNew` if say the user switches from using `http` <-> `https`?

davidji99:
Hey @mars in reviewing this PR, I'd like to know about the use case for the `heroku_slug` resource when someone doesn't specify a `file_path` or `file_url`. It doesn't seem `heroku_slug` would be of much use if they uploaded the slug outside of terraform?

For example, if someone follows the advice listed here in the doc:
```
...to the resulting computed `blob.method` + `blob.url` by some other means, otherwise app release will fail with _Compiled slug couldn't be found_
```
that person would have already done a `POST /apps/{app_id_or_name}/slugs` in order to get the `blob.url` for upload externally, which is essentially what heroku_slug resource would do without a file path or url. I would suggest that we modify this resource to stay true to the docs as a resource that "Provides the ability to create **& upload** a slug (archive of executable code) to an app, making it possible to launch apps directly from a Terraform config".

heroku/resource_heroku_slug.go
mars: Nice! `ConflictsWith` is much better. Will do.

mars:
Hi @davidji99,

> It doesn't seem heroku_slug would be of much use if they uploaded the slug outside of terraform?

I totally agree. That alternative behavior was a result of my incremental development process, and it seems to have very little utility, instead adding adds unnecessary complexity.

I will revise to require either `file_path` or `file_url` and update/simplify the docs accordingly.

website/docs/r/slug.html.markdown
```diff
@@ -17,10 +17,10 @@ This resource supports uploading a pre-generated archive file of executable code
...
-* When `file_path` is not specified, then the [slug archive must be uploaded](https://devcenter.heroku.com/articles/platform-api-deploying-slugs#publish-to-the-platform) to the resulting computed `blob.method` + `blob.url` by some other means, otherwise app release will fail with _Compiled slug couldn't be found_
+* When `file_url` or `file_path` is specified, the referenced file must contain a slug archive of executable code and must follow the prescribed layout from [Create slug archive](https://devcenter.heroku.com/articles/platform-api-deploying-slugs#create-slug-archive) in the Heroku Dev Center (nested within an `./app` directory)
+* When neither `file_url` nor `file_path` is specified, then the [slug archive must be uploaded](https://devcenter.heroku.com/articles/platform-api-deploying-slugs#publish-to-the-platform) to the resulting computed `blob.method` + `blob.url` by some other means, otherwise app release will fail with _Compiled slug couldn't be found_
```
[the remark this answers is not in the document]
mars: This part of the docs have been deleted, because now the resource always uploads the slug.

heroku/resource_heroku_slug.go
mars: I'm not convinced that `DiffSuppressFunc` for changing the protocol is a good idea. If we're going to check the protocol, I'd rather just prevent insecure `http` URLs.

mars:
All comments addressed with the last commit. Also rebased on current master.

heroku/resource_heroku_slug.go
davidji99: 'If we're going to check the protocol, I'd rather just prevent insecure http URLs.' <-- That might force people to use `https`, which seems like a good idea?

mars: `file_url` is now validated to start with `https://` 🔒 👍

heroku/resource_heroku_slug.go
```diff
@@ -40,6 +42,15 @@ func resourceHerokuSlug() *schema.Resource {
 				ForceNew: true,
```
davidji99: Do we also need a `ConflictsWith: []string{"file_url"}` here?

davidji99:
Looks good @mars. Just had one more question and I think this should be good to go.

heroku/resource_heroku_slug.go
davidji99: Sounds good!

PR #142, opened 2018-11-13
Add provider configuration for backend

talbright:
This is an implementation proposal based on the discussion in #111, whereby we allow additional configuration to be passed into the provider to help alleviate issues with cedars backend eventual consistency with regards to app, spaces, and domain creation.

```hcl
provider "heroku" {
  "delays" : {
     post_app_create_delay = 15
     post_domain_create_delay = 12
     post_space_create_delay = 20  
  }
}
```

Naturally a user *can* do something like: 

```hcl
resource "heroku_app" "default" {
  provisioner "local-exec" {
    command = "sleep 15"
  }
}
```

But this ends up being littered across users HCL files.  When this issue is resolved in the Heroku platform this can be removed or become a no-op.

### DIFF
--- a/heroku/provider.go
+++ b/heroku/provider.go
@@ -44,7 +44,7 @@ func Provider() terraform.ResourceProvider {
 				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("HEROKU_API_URL", heroku.DefaultURL),
 			},
-			"api": {
+			"delays": {
 				Type:     schema.TypeList,
 				MaxItems: 1,
 				Optional: true,
@@ -142,7 +142,7 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		config.APIKey = apiKey.(string)
 	}
 
-	err = applyAPIConfig(d, &config)
+	err = applyDelayConfig(d, &config)
 
 	log.Println("[INFO] Initializing Heroku client")
 	if err := config.loadAndInitialize(); err != nil {
@@ -152,8 +152,8 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	return &config, nil
 }
 
-func applyAPIConfig(d *schema.ResourceData, config *Config) error {
-	if v, ok := d.GetOk("api"); ok {
+func applyDelayConfig(d *schema.ResourceData, config *Config) error {
+	if v, ok := d.GetOk("delays"); ok {
 		vL := v.([]interface{})
 		if len(vL) > 1 {
 			return fmt.Errorf("Provider configuration error: only 1 api config is permitted")
--- a/heroku/provider.go
+++ b/heroku/provider.go
@@ -1,22 +1,14 @@
 package heroku
 
 import (
-	"encoding/json"
 	"fmt"
 	"log"
-	"net/http"
-	"net/url"
 	"strings"
 
-	"os"
-	"runtime"
-
-	"github.com/bgentry/go-netrc/netrc"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
 	"github.com/hashicorp/terraform/terraform"
 	heroku "github.com/heroku/heroku-go/v3"
-	"github.com/mitchellh/go-homedir"
 )
 
 // Provider returns a terraform.ResourceProvider.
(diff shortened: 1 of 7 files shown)


mars:
@talbright Making this a provider config sounds great 👍 

Will the actual delay be implemented for the `heroku_app` resource in this same PR?

Also, would you consider making the default > 0 to help avoid folks needing to know about this at all?

talbright:
> @talbright Making this a provider config sounds great 👍
> 
> Will the actual delay be implemented for the `heroku_app` resource in this same PR?
💯 Just wanted to get some feedback first. I will also add delays (and config) for space and domain creation in this PR.

> Also, would you consider making the default > 0 to help avoid folks needing to know about this at all?
Agreed, I was thinking the same last night. What do you think a sane default value should be?

heroku/provider.go
```diff
@@ -35,6 +37,20 @@ func Provider() terraform.ResourceProvider {
 				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("HEROKU_HEADERS", nil),
 			},
+			"api": {
```
davidji99: Thinking out loud here - do we want to organize these options by resource? For example, `app_opts`, `domain_opts` which then breaks down further to the specific options like the one you suggested below? I guess `api` is further up category-wise but figure some more granular naming for organization sake might be worth thinking about in the long term.

heroku/provider.go
```diff
@@ -35,6 +37,20 @@ func Provider() terraform.ResourceProvider {
...
+				Type:     schema.TypeList,
```
davidji99: Curious, why `schema.TypeList` over `schema.TypeMap`? I'd imagine when we have more options, it would be:  
```
api: {
  "option_1" : value_1,
 "option_2": value_2,
}

**Comment by davidji99** [comment]
```
Also, would you consider making the default > 0 to help avoid folks needing to know about this at all?
```

I would suggest that we consider opt-in flow of these options instead of opt-out to preserve current behavior. Or are you guys suggesting that a non-zero value for `post_app_create_delay` is a necessity these days?

**Comment by talbright** [comment]
> ```
> Also, would you consider making the default > 0 to help avoid folks needing to know about this at all?
> ```
> I would suggest that we consider opt-in flow of these options instead of opt-out to preserve current behavior. Or are you guys suggesting that a non-zero value for `post_app_create_delay` is a necessity these days?

There's a couple of advantages I see to opt-out:
* ideally we catch those occasional errors we've seen repeatedly which is a better experience for our users, and requires them to do nothing
* when/if its fixed in the back-end nothing needs to be changed by users 

I admit though that both of these options assume a reasonable default value for the delays which is a bit of a moving target.

**Comment by mars** [comment]
> Or are you guys suggesting that a non-zero value for `post_app_create_delay` is a necessity these days?

Which is more valuable, Performance 🆚 Elegance. Either we,

**Set default delay greater than 0.** Inhibit the provider's performance so every `heroku_app` created take an extra N-seconds. (Hopefully less than fifteen-seconds, but I do not have evidence to inform the wait time.)

*or*

**Set default delay to 0.** Folks continue to intermittently experience this issue, and then they must figure out how to fix it by tweaking  `post_app_create_delay`.

-----

I vote for Elegance, but would crank down the default wait time as much as possible. Perhaps we investigate the nominal, expected delay with Heroku runtime.

File: heroku/provider.go (hunk 2 above)
Comment (talbright) [reply heroku/provider.go]: Well truth be told, I'm still working through what are the best types to use for which scenarios in more complex schemas (`TypeList` vs `TypeSet`). 

This does however allow one to access the options themselves as a map, see https://github.com/terraform-providers/terraform-provider-heroku/blob/2d54f078300247c0916aaa4c5d66e30bea595285/heroku/provider.go#L134-L138

**Comment by mars** [comment]
I just spoke with a Heroku Runtime engineering manager, and it sounds like reasonable app creation wait time would be:

* Common Runtime: 5-seconds
* Private Spaces: longer than Common Runtime, but unknown
* Shield Private Spaces: longer than Private Spaces, but unknown

They reflected that the right way to check if it's ready (instead of a standard wait delay) is to simply try fetching the app. If 404 is returned it's not ready.

I'm not sure if this is new information for you all. It was for me!

**Comment by talbright** [comment]
> I just spoke with a Heroku Runtime engineering manager, and it sounds like reasonable app creation wait time would be:
> 
> * Common Runtime: 5-seconds
> * Private Spaces: longer than Common Runtime, but unknown
> * Shield Private Spaces: longer than Private Spaces, but unknown
✅ its a start

> They reflected that the right way to check if it's ready (instead of a standard wait delay) is to simply try fetching the app. If 404 is returned it's not ready.
> 
This is the source of the problem for both apps and spaces -- the API returns the resource but its not actually ready for other actions to be performed on it.

In the case of spaces we even use a RefreshFunc to poll: https://github.com/terraform-providers/terraform-provider-heroku/blob/master/heroku/resource_heroku_space.go#L107. Still, even after getting a space resource response from the back-end and adding it to the state, problems ensue. 

For apps the issue has always been that the API returns the resource right away, but in truth its not quite ready for other actions to be performed on it (even though we successfully add it into the state), which causes downstream errors for other resources. 

You could hit up @bernerdschaefer and @sclasen, I chatted with them both a while back, they were aware of the issue.

**Comment by mars** [comment]
@talbright thanks for the background. Clearly I got "hopeful" information instead of "accurate" 😬

**Comment by talbright** [comment]
@davidji99 remember, this isn't about how long it takes the back-end to create an app in common, ps, or shield spaces. It's an eventual consistency problem. There could well be different timings for this eventual consistency, but to keep things simple for now lets just use one value to cover all three. If it turns out to be insufficient, I can add more.

File: heroku/provider.go (hunk 1 above)
Comment (talbright) [reply heroku/provider.go]: I'm 50/50 here between just keeping it simple and creating per resource blocks.

**Comment by talbright** [comment]
Changed my mind, I think 5s a better default.

**Comment by mars** [comment]
I'm finding that the default is not actually taking effect.

I have to set it explicitly to get the delay:

```hcl
provider "heroku" {
  "delays" = {
     "post_app_create_delay" = 10
  }
}
```

talbright:
🤔 ok will take a look

talbright:
@mars fixed and tested locally.

Do you still think we should stick with 10s, given you probably ran into a the bug with the value being set properly?

PR #145, opened 2018-11-16
Terraform 0.12

mars:
Working on [a different PR](https://github.com/hashicorp/terraform/pull/19070) to 🤞 land in the next major version of Terraform, 0.12. The new version changes the internal API for plugins, forcing a recompile of provider plugins with the new version.

> This release introduces new wire protocols for provider and provisioner plugins and a new automatic installation method for provider plugins. At the time of release there are no official plugin releases compatible with these new protocols and so automatic provider installation with terraform init is not functional. Instead, the v0.12.0-alpha1 distribution archives contain bundled experimental provider builds for use with the alpha.
> …
> Provider developers may wish to try building their plugins against the v0.12-alpha1 tag of Terraform Core to use them with this build. We cannot yet promise that all providers will be buildable in this way and that they will work flawlessly after building. Official releases of all HashiCorp-hosted providers compatible with v0.12 will follow at some point before v0.12.0 final.

–[Terraform CHANGELOG](https://github.com/hashicorp/terraform/blob/master/CHANGELOG.md#0120-alpha1-october-19-2018)

Therefore, trying to build this Heroku provider against Terraform master. Here's what I've got so far. Will comment with progress…


mars:
Will reopen for the currently stabilizing terraform@master.

PR #149, opened 2018-11-21
Build resource

mars:
Go from source code to an app release with a single Terraform resource!

***Deploy to Heroku** for Terraform.*

See [the docs](https://github.com/mars/terraform-provider-heroku/blob/build-resource/website/docs/r/build.html.markdown) for the details.


talbright:
This looks promising. I'll be out for the holidays and at a conference next week, so I won't have much time until I get back to dig into this. However, I had a couple of questions/thoughts:

* Generating the tarball is still out-of-band from Terraform. What sort of workflows or patterns do you envision here?
* Why does it have to be a tarball? Can I just point it to a path and have the provider tarball it for me?
* Would there be benefits for this to work more directly with `git push heroku master`

mars:
Great to hear @talbright 😄 Happy Thanksgiving & then a fantastic big conf to you!

> * Generating the tarball is still out-of-band from Terraform. What sort of workflows or patterns do you envision here?

For Source URLs, GitHub auto-generates tarballs for releases & branches. So that's super convenient. Set the desired GitHub release tag or branch as a TF variable, and use it in the Build resource's `source.url` & `source.version` values.

> * Why does it have to be a tarball? Can I just point it to a path and have the provider tarball it for me?

Supporting a `source.path` **directory**, in addition to a **tarball**, is a Great Idea 💯  Tarballs are the basis of Build & Source APIs. A little branching logic upon create and generating the tarball from Go should be all it takes.

This will enable a local app development workflow, swiftly iterating through `terraform apply` runs with uncommitted source code & Terraform configurations in a local working directory.

> * Would there be benefits for this to work more directly with `git push heroku master`

`git push heroku master` is a different way to do the same thing. This resource uses Heroku Platform [Build & Source APIs](https://devcenter.heroku.com/articles/build-and-release-using-the-api), *instead* of the Heroku App git remote.

Perhaps the `git push heroku master` metaphor is confusing, because git is not part of this Build resource, except for the [GitHub Source URL guidance](https://github.com/mars/terraform-provider-heroku/blob/build-resource/website/docs/r/build.html.markdown#github-urls). Currently, Heroku Platform API does not support much git or GitHub integration, only an [App `git_url`](https://devcenter.heroku.com/articles/platform-api-reference#app) attribute.

mars:
Another Heroku git fact:

When deployed using the Builds API, a Heroku app's git remote is an empty repo.

So, apps deployed through this new resource do not have any data in Heroku git. This is the same behavior as apps deployed with the official, purple **Deploy to Heroku** app setup button.

mars:
`source.path` now supports a directory. Just point the terraform config at a local working directory to deploy it:

```hcl
  name = "${var.example_app_name}"
  region = "us"
}

resource "heroku_build" "example" {
  app = "${heroku_app.example.id}"
  
  source = {
    path    = "../example-microservice-api"
  }
}

resource "heroku_formation" "example" {
  app        = "${heroku_app.example.id}"
  type       = "web"
  quantity   = 1
  size       = "Standard-1x"
  depends_on = ["heroku_build.example"]
}
```

mars:
Happy December 🎄 @talbright & all. Would you have a chance to review this soon?

talbright:
🦌gimme a couple of days to get to it

heroku/resource_heroku_build.go
```diff
@@ -0,0 +1,506 @@
...
+	tarballPath := fmt.Sprintf("source-%s.tar.gz", newUuid)
+	err = tarinator.Tarinate(sourcePaths, tarballPath)
+	if err != nil {
+		return "", fmt.Errorf("Error generating build source tarball %s of %s: %s", tarballPath, path, err)
+	}
+	return tarballPath, nil
+}
+
+// Returns a resource.StateRefreshFunc that is used to watch a Build.
+func BuildStateRefreshFunc(client *heroku.Service, app, id string) resource.StateRefreshFunc {
```
talbright: ✅ ideally we figure out a way to stream the build output, on the other hand this is pragmatic.

heroku/resource_heroku_build.go
```diff
@@ -0,0 +1,506 @@
...
+	return
+}
+
+func generateSourceTarball(path string) (filePath string, err error) {
+	sourcePaths := []string{path}
+	newUuid, err := uuid.GenerateUUID()
+	if err != nil {
+		return "", err
+	}
+	tarballPath := fmt.Sprintf("source-%s.tar.gz", newUuid)
```
talbright: You could just use `filePath` here: 

```
filePath = fmt.Sprintf("source-%s.tar.gz", newUuid)
```

heroku/resource_heroku_build.go
```diff
@@ -0,0 +1,506 @@
...
+	err = tarinator.Tarinate(sourcePaths, tarballPath)
+	if err != nil {
+		return "", fmt.Errorf("Error generating build source tarball %s of %s: %s", tarballPath, path, err)
+	}
+	return tarballPath, nil
```
talbright: If you used `filepath`, you could optionally just use `return` like you did in the `validateSourceUrl` function.

heroku/resource_heroku_build.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+
+	setBuildState(d, build, app)
+
+	return nil
+}
+
+// A no-op method as there is no DELETE build in Heroku Platform API.
+func resourceHerokuBuildDelete(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[INFO] There is no DELETE for build resource so this is a no-op. Build will be removed from state.")
```
talbright: Don't we need:

```
d.SetId("")
```

heroku/resource_heroku_build.go
```diff
@@ -0,0 +1,506 @@
...
+
+	d.SetId(build.ID)
+	setBuildState(d, build, app)
+
+	log.Printf("[INFO] Created build ID: %s", d.Id())
+
+	return nil
+}
+
+func resourceHerokuBuildRead(d *schema.ResourceData, meta interface{}) error {
```
talbright: ✅

Gopkg.toml
```diff
@@ -48,3 +48,7 @@
 [[constraint]]
   name = "github.com/stretchr/testify"
   version = "1.2.2"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/verybluebot/tarinator-go"
```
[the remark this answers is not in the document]
mars: Thank you. I totally agree. I was reluctant to bring in a new dep, but Go's stdlib tar does not handle directories & recursion at all, forcing a lot of new code to do this seemingly simple thing.

website/docs/r/build.html.markdown
```diff
@@ -0,0 +1,149 @@
...
+
+```hcl
+resource "heroku_app" "foobar" {
+    name   = "foobar"
+    region = "us"
+}
+
+resource "heroku_build" "foobar" {
+  app        = "${heroku_app.foobar.id}"
+  buildpacks = ["https://github.com/mars/create-react-app-buildpack"]
```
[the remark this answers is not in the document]
mars: Each Build overrides whatever buildpacks were previously set, whether originally set on the app or by a build. I just verified this using a simple Terraform config, snippet:

```hcl
  name = "${var.example_app_name}"
  region = "us"
  buildpacks = ["https://github.com/heroku/heroku-buildpack-scala"]
}

resource "heroku_build" "example" {
  app = "${heroku_app.example.id}"
  buildpacks = ["https://github.com/heroku/heroku-buildpack-ruby"]

  source = {
    path = "app/"
  }
}
```

Changing the `heroku_build.example` `buildpacks` & re-applying the config overrides that last one set.

website/docs/r/build.html.markdown
```diff
@@ -0,0 +1,149 @@
...
+  * `version` - Use to track what version of your source originated this build. If you are creating builds from git-versioned source code, for example, the commit hash, or release tag would be a good value to use for the version parameter.
+
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `uuid` - The ID of the build
+* `output_stream_url` - URL that [streams the log output from the build](https://devcenter.heroku.com/articles/build-and-release-using-the-api#streaming-build-output)
+* `release_id` - The Heroku app release created with a build's slug
```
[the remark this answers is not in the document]
mars: The Build `id`/`uuid` is different than the `release_id`

For example, here's the state of the build I just used to test the buildpacks setting precedence:

```
heroku_build.example:
  id = e624119b-4016-47e1-bd54-d3c4a8cf344a
  app = mars-terraforming-example
  buildpacks.# = 1
  buildpacks.0 = https://github.com/heroku/heroku-buildpack-ruby
  local_checksum = SHA256:2000fcb8622d99766b177cdadc4acfeac61625657f013a561d63a2e28b510e48
  output_stream_url = https://build-output.heroku.com/streams/xxxxx
  release_id = 487be925-57e3-401c-b186-6fadd9848429
  slug_id = 303aa988-a46f-4e83-928e-648cb039baab
  source.% = 1
  source.path = app/
  stack = heroku-18
  status = succeeded
  user.% = 2
  user.email = mars@heroku.com
  user.id = d5dfd01f-4252-40c4-8620-e1f28a7d27fc
  uuid = e624119b-4016-47e1-bd54-d3c4a8cf344a
```

Each **Build** creates a **Slug** & a **Release**.

heroku/resource_heroku_build.go
mars: > This operation should never update any state on the resource. It is not necessary to call d.SetId(""), since any non-error return value assumes the resource was deleted successfully.

–https://www.terraform.io/docs/extend/writing-custom-providers.html#implementing-destroy

heroku/resource_heroku_build.go
mars: I've pondered this as well 😇

Due to our provider convention to hide underlying log output during Terraform runs and the complexity (an assumption from a still new Gopher here) of such a streaming log implementation, I opted to just spit out the Build's log URL when it fails.

Otherwise the build log can be viewed by the Terraform consumer with the `heroku_build` `output_stream_url` that's stored in state.

mars:
Thanks @talbright for the careful review 🙇‍♂️ 

I believe I've addressed all the comments. Barring further challenges, I think this is ready to go 🏁

website/docs/r/build.html.markdown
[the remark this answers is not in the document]
talbright: Got it...my misunderstanding then

website/docs/r/build.html.markdown
```diff
@@ -0,0 +1,149 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `app` - (Required) The ID of the Heroku app
+* `buildpacks` - List of buildpack registry names and/or GitHub URLs
+* `source` - (Required) A block that specifies the source code to build & release:
+  * `checksum` - Hash of the source archive for verifying its integrity, auto-generated when `source.path` is set, `SHA256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`
```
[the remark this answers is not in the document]
talbright: That makes sense then.

heroku/resource_heroku_build.go
talbright: I see, good to know. That seems to be a bit of cargo cult coding in many resources we could get rid of.

talbright:
> > Can you address the build failures?
> 
> After rebasing, the errors revealed themselves locally, so fixed, now passing.
> 
> But, `TF_ACC` is not set in Travis CI, so these tests are not really running for this new resource. It is passing locally:
> 
👍

PR #151, opened 2018-12-10
Allow setting URL

tt:
Heroku has multiple environments and it would be great if we could use Terraform to manage resources in these.

I decided to not document usage in the website/docs/index.html.markdown file because of the limited utility. Let me know if I should.

### DIFF
--- a/heroku/config.go
+++ b/heroku/config.go
@@ -11,6 +11,7 @@ type Config struct {
 	Email   string
 	APIKey  string
 	Headers http.Header
+	URL     string
 
 	Api *heroku.Service
 }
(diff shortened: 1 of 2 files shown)


tt:
@talbright, I rebased this branch.

The intent of this is only making the endpoint configurable to have parity with our other tooling. It would indeed be confusing if we also began including resources that aren't available to users and the policy is a great forcing function for us to ship features to general availability.

davidji99:
@tt (In a new PR), Perhaps we should document that this is for Heroku-internal use only so people don't wonder about what `URL` is used for?

talbright:
@davidji99 I think its better left undocumented so as not to cause confusion. The only way a user would know about this is to look in the source code.

PR #158, opened 2018-12-16
add provider delay config for app delete

talbright:
This is a continuation of #142 to add an additional `post_app_delete_delay` to address #123.

```hcl
provider "heroku" {
  "delays"  {
     post_app_delete_delay = 10  
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,12 @@
 ## 1.7.1 (Unreleased)
+
+FEATURES:
+
+IMPROVEMENTS:
+* Add configurable delay for post app deletion [#158](https://github.com/terraform-providers/terraform-provider-heroku/pull/158)
+
+BUG FIXES:
+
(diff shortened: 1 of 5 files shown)


talbright:
I still haven't been able to recreate the issue. Probably I need to create real dynos instead of just the apps per #159 . Might be a good chance to give the fancy new build resource a try.

talbright:
Looks like I forgot to come back to this! I was never able to reproduce the issue, do you think we should close this or put it in anyways since it was reported in #123?

talbright:
I'll close this for now, if it comes up again we can raise it back from the 💀

mars:
I've Terraform destroyed a Space with apps in it sooo many times, never encountered an eventual consistency issue on app delete. I wonder if the initial bug reported was a manifestation of some underlying Private Spaces behavior or bug that has since been changed. ANyway. Thanks for cleaning this up @talbright !

PR #161, opened 2018-12-19
Identify Terraform API requests via user-agent header

mars:
In order to facilitate Heroku's desire to observe how tools such as Terraform are used with the platform, set a custom "User-Agent" value for this Terraform provider.

Currently, requests to the Heroku API use the [default HTTP "User-Agent" header from heroku-go](https://github.com/heroku/heroku-go/blob/master/v3/heroku.go#L29), such as:

```
heroku/v3 (linux; amd64)
```
 
This pull request simply will set a custom value to self-identify, and it includes heroku-go's original value:

```
heroku/v3 (linux; amd64) terraform-provider-heroku/v1.7.1
```

Example provider request header:

```
GET /apps/tftest-9yz1nyjtsl/builds/093ad1ad-3f8c-4f65-9f6e-5c73d488753e HTTP/1.1
Host: api.heroku.com
User-Agent: heroku/v3 (darwin; amd64) terraform-provider-heroku/v1.7.1
Accept: application/vnd.heroku+json; version=3
Authorization: Basic xxxxx
Request-Id: 1f0c1bfc-2da7-4b6f-b643-246d7422b66a
Accept-Encoding: gzip
```

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -14,7 +14,7 @@ test: fmtcheck
 		xargs -t -n4 go test -v $(TESTARGS) -timeout=30s -parallel=4
 
 testacc: fmtcheck
-	TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout 120m -ldflags="-X=github.com/terraform-providers/terraform-provider-heroku/version.ProviderVersion=acc"
+	TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout 120m -ldflags="-X=github.com/terraform-providers/terraform-provider-heroku/version.ProviderVersion=test"
 
 vet:
 	@echo "go vet ."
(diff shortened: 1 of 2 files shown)


mars:
I just realized the RFC says "most significant" first, not "most specific", will reorder as you suggested @talbright 

Also, when I have a bit more time, will get the plugin version number as suggested.

mars:
Thanks @talbright for making `version.ProviderVersion` available!

The User-Agent header has been updated to include the value, here seen as `acc` when running tests:

```
User-Agent: heroku/v3 (darwin; amd64) terraform-provider-heroku/acc
```

The `acc` will instead be the current version when Hashicorp build's the provider for release.

So, this should be ready to go 🏁

mars:
Any idea why `acc` is the test value, instead of something obvious (to me) like `test`?

talbright:
> Any idea why `acc` is the test value, instead of something obvious (to me) like `test`?

Most likely to follow the `acc` nomenclature present in the TF code base (presumably to emphasize acceptance test over unit test.) I'm not against it being changed to `test` though...

mars:
Okay, changed the test version to `test`:

```
User-Agent: heroku/v3 (darwin; amd64) terraform-provider-heroku/test
```

So with that, I'll be merging :shipit:

PR #163, opened 2018-12-25
New Heroku App attribute: sensitive_config_vars

### LINKED ISSUE #72: Ability to hide config vars from Plan Output

davidji99:
Resolves #72

Continuing from #101, this issue/PR will introduce a `sensitive_config_vars` attribute to the `heroku_app` resource. This will allow users to add in passwords/tokens/secrets into their .tf files without risking it being exposed in CI builds.

Sample HCL:
```
  name   = "foobar"
  region = "us"

  config_vars {
    RAILS_ENV = "staging"
  }
  
  sensitive_config_vars {
    DATABASE_URL = "postgres://user:password@some-host.com:1234"
  }
}
```

### DIFF
--- a/heroku/resource_heroku_app.go
+++ b/heroku/resource_heroku_app.go
@@ -166,6 +166,16 @@ func resourceHerokuApp() *schema.Resource {
 				},
 			},
 
+			"sensitive_config_vars": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type:      schema.TypeMap,
+					Sensitive: true,
+				},
+			},
+
 			"all_config_vars": {
 				Type:     schema.TypeMap,
 				Computed: true,
@@ -404,6 +414,9 @@ func resourceHerokuAppRead(d *schema.ResourceData, meta interface{}) error {
 	care := make(map[string]struct{})
 	configVars := make(map[string]string)
 
+	careSensitive := make(map[string]struct{})
+	sensitiveConfigVars := make(map[string]string)
+
 	// Only track buildpacks when set in the configuration.
 	_, buildpacksConfigured := d.GetOk("buildpacks")
 
@@ -433,9 +446,31 @@ func resourceHerokuAppRead(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
+	for _, v := range d.Get("sensitive_config_vars").([]interface{}) {
+		// Protect against panic on type cast for a nil-length array or map
+		n, ok := v.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		for k := range n {
+			careSensitive[k] = struct{}{}
+		}
+	}
+
+	for k, v := range app.Vars {
+		if _, ok := careSensitive[k]; ok {
+			sensitiveConfigVars[k] = v
+		}
+	}
+
 	var configVarsValue []map[string]string
 	if len(configVars) > 0 {
 		configVarsValue = []map[string]string{configVars}
+	}
+
+	var sensitiveConfigVarsValue []map[string]string
+	if len(sensitiveConfigVars) > 0 {
+		sensitiveConfigVarsValue = []map[string]string{sensitiveConfigVars}
 	}
 
 	if buildpacksConfigured {
@@ -444,6 +479,10 @@ func resourceHerokuAppRead(d *schema.ResourceData, meta interface{}) error {
 
 	if err := d.Set("config_vars", configVarsValue); err != nil {
 		log.Printf("[WARN] Error setting config vars: %s", err)
+	}
+
+	if err := d.Set("sensitive_config_vars", sensitiveConfigVarsValue); err != nil {
+		log.Printf("[WARN] Error setting sensitive config vars: %s", err)
 	}
 
 	if err := d.Set("all_config_vars", app.Vars); err != nil {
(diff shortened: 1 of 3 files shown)


davidji99:
Say I have the following terraform config originally:
```
  name = "foobar-app-name"
  region = "us"

  organization {
    name = "sfdc-ohana-terraform-sandbox"
  }

  config_vars = {
    RAILS_ENV = "production"
    DATABASE_PASSWORD = "fakepassword"
  }
}
```

With the new introduction of the `sensitive_config_vars` attribute, I now want to move `DATABASE_PASSWORD` to this attribute. So the terraform config becomes:
```
  name = "sfdc-terraform-provider-tester"
  region = "us"

  organization {
    name = "sfdc-ohana-terraform-sandbox"
  }

  config_vars = {
    RAILS_ENV = "production"
  }

  sensitive_config_vars = {
    DATABASE_PASSWORD = "fakepassword"
  }
```

The terraform plan diff would look like this:
```
Terraform will perform the following actions:

  ~ heroku_app.foobar
      config_vars.0.%:                           "2" => "1"
      config_vars.0.DATABASE_PASSWORD:           "fakepassword" => ""
      sensitive_config_vars.#:                   "0" => "1"
      sensitive_config_vars.0.%:                 "0" => "1"
      sensitive_config_vars.0.DATABASE_PASSWORD: <sensitive> => <sensitive> (attribute changed)
```
The immediate problem I see here is that `DATABASE_PASSWORD`'s value is in the diff so the config var needs to be rotated post `terraform apply`. I'm not sure this would be an optimal workflow. Plus, there is a bug in terraform-core in that when you unset `config_vars` (assuming that attribute was marked as `Sensitive`), you'll see the value plaintexted in the diff.

What do you guys think @talbright @mars?

talbright:
> The immediate problem I see here is that `DATABASE_PASSWORD`'s value is in the diff so the config var needs to be rotated post `terraform apply`. I'm not sure this would be an optimal workflow. Plus, there is a bug in terraform-core in that when you unset `config_vars` (assuming that attribute was marked as `Sensitive`), you'll see the value plaintexted in the diff.
> 
I appreciate your attention to detail and considering all scenarios. 

For the migration concern, this seems like something we can ignore. Consider that if the user has `DATABASE_PASSWORD` in `config_vars`, that value has _already been exposed in diffs_. Your changes won't expose a new problem, it fixes future ones.

Regarding the bug in terraform-core, is that issue being tracked? Can you send us the link? This issue shouldn't stop your work though. Call it out loudly and clearly in the docs. In the meantime, we have enterprise support. I'm willing to help escalate the problem through that channel if needed to see if we can get it prioritized.

davidji99:
Thanks Trent.

> For the migration concern, this seems like something we can ignore. Consider that if the user has DATABASE_PASSWORD in config_vars, that value has already been exposed in diffs. Your changes won't expose a new problem, it fixes future ones.

Make senses. Works for me.

> Regarding the bug in terraform-core, is that issue being tracked? Can you send us the link? This issue shouldn't stop your work though. Call it out loudly and clearly in the docs. In the meantime, we have enterprise support. I'm willing to help escalate the problem through that channel if needed to see if we can get it prioritized.

Let me find it. If not, I'll create an issue on terraform-core repository. And yeah, it shouldn't stop work on this PR.

davidji99:
@talbright @bernerdschaefer @mars - I think this PR is ready to go. Please review.

joslynesser:
Thanks for getting this one in @davidji99! 🍻

mars:
I see that swapping vars is [tested](https://github.com/terraform-providers/terraform-provider-heroku/pull/163/files#diff-284c0587965cd4ad5c279fd2863f1ec4R887), but the case I encountered is a complete replacement of the vars attribute:

```hcl
  name = "${var.example_app_name}"
  region = "us"

  config_vars = {
    BIG_OLD_SECRET = "for your eyes only"
  }
}
```

→

```hcl
  name = "${var.example_app_name}"
  region = "us"

  sensitive_config_vars = {
    BIG_OLD_SECRET = "for your eyes only"
  }
}
```

mars:
Thanks to @davidji99 for giving me a bit more context, I see that the behavior of removing `config_vars` block is a noop.

So, the correct way to swap vars between the `config_vars` & `sensitive_config_vars` attributes is to always leave both blocks in the configuration, even if empty:

```hcl
  name = "${var.example_app_name}"
  region = "us"

  config_vars = {
    BIG_OLD_SECRET = "for your eyes only"
  }

  sensitive_config_vars = {
  }
}
```

→

```hcl
  name = "${var.example_app_name}"
  region = "us"

  config_vars = {
  }

  sensitive_config_vars = {
    BIG_OLD_SECRET = "for your eyes only"
  }
}
```

Migrating vars works fine, then ✅

So, my request is to document this as part of the `heroku_app` resource. Maybe a "Note" that says something like:

> Removing an entire `config_vars` or `sensitive_config_vars` block from a configuration will not actually remove the vars they declare. This is important when migrating all `config_vars` to `sensitive_config_vars`. Leave the empty `config_vars` or `sensitive_config_vars` block in place to remove previously defined vars.

davidji99:
@mars added the 'Note' as you requested.

PR #165, opened 2019-01-18
Switch to go modules

appilon:
Looking ahead to Terraform 0.12 switching to go modules for dependency management is advised. This PR will track the dependencies with modules however we still `vendor/` and force it with `GOFLAGS=-mod=vendor` to ensure CI remains unaffected. TeamCity had only 1 new failure (however it appears to be a flaky test).


mars:
Hi @appilon , thank you for proposing this! ☺️

I've checked out this branch and can `make build` & use the provider no problem, but we definitely should update the [dependency docs](https://github.com/terraform-providers/terraform-provider-heroku/blob/master/README.md#user-content-updating-or-adding-dependencies) before merging. I would recommend those updates, but do not know what they should say. If anyoneone else can contribute to the right way to update or add deps, please do.

appilon:
Hello @mars I am back from vacation and have updated the docs. Unfortunately it appears a PR was merged that updates a dependency with dep. I will need to restart migration!

mars:
Sorry about that @appilon 😬 I'll shepherd it through to merge next time 🙏

PR #182, opened 2019-02-26
Start standardizing on either google/uuid or hashicorp/go-uuid

freeformz:
This removes the dependency on satori/uuid in exchange for
google/uuid.

pborman/uuid is now a wrapper around google/uuid, so I expect users of
it to fall out to google/uuid over time.

I also made a few small tweaks the to validateUUID function to include
a case where the interface{} can't be converted to a string.


davidji99:
Closing this PR as #228 has been merged.

Issue #124, opened 2018-09-13
Renaming an App breaks many things


mars:
Because of the prevalence of using App.Name identifiers throughout the Heroku resources, if an app is renamed after the initial `terraform apply`, then state may no longer refresh without error.

Here's an example error with the Slug resource after an app rename:

> Error: Error refreshing state: 1 error(s) occurred:
>
>* heroku_slug.wasabi: 1 error(s) occurred:
>
>* heroku_slug.wasabi: heroku_slug.wasabi: Error retrieving slug: Get https://api.heroku.com/apps/mars-terra-micro-wasabi-int/slugs/bb8579ca-a052-4fba-9a68-d250b0d115d5: Couldn't find that app.

*The config drifts* ⛵️☠️

Based on evidence in #93, clearly this is a known challenge.

I checked many of the resources and found two classes of this problem:

1. **AppRelease**, **Formation**, **TeamCollaborator**, & **Slug** seem to always have this problem
2. **AddOn**, **AppFeature**, **Cert**, **Domain**, **Drain** (potentially more) have this problem if they are created using an App Name instead of its UUID.

Possible solution: I'd like to try migrating all the App.Name/AppName usage to App.ID (UUID), but this touches a lot of resources.

Before attempting this, I'd like to hear some feedback from you all about other possible solutions, problems this would cause, or lurking problems of which you might be aware.

talbright:
As a matter of floor procedure I suggest we keep the convo in #93, although the title is different the end conclusion proposed is the same as here.

PR #183, opened 2019-03-04
New Resource: Heroku Config & Heroku App Config Association

davidji99:
## TODO:
- [x] Write tests
- [x] Add Docs

## Questions to ponder:
- [x] Should remain `heroku_config_association` to `heroku_config_app_association`?
- [x] Should remove `name` attribute from `heroku_config`?

## Example HCL
```
resource "heroku_app" "foobar"  {}

resource "heroku_config" "foobar-configs" {
    vars = {
        RAILS_ENV="production"
    }
   
    sensitive_vars = {
          PRIVATE_KEY="wouldn't_you_like_to_know"
    }
}

resource "heroku_app_config_association" "foobar" {
    app_id = "${heroku_app.foobar.id}"
    config_vars = "${heroku_config.foobar-configs.vars}"
    sensitive_config_vars = "${heroku_config.foobar-configs.sensitive_vars}"
}
```

The naming is up for debate. Just starting this PR for discussion.

## Reason
The purpose of the proposed config vars is for users to set heroku app config vars from a separate repo. Imagine a git repo called `kafka-addon` that holds the HCL for configuring a heroku kafka addon instance. With the two proposed resources, a user can reference the app IDs of all the apps that use kafka and set the kafka vars on each app without needing to making multiple changes to each individual apps or services' repository. Taking the kafka example further, imagine this scenario:

- User has separate `git` repos for various services in a microservice layout.
- User creates a kafka terraform configuration in its own repository with blue/green support.
- User builds out new clusters and update state.
- User has to deploy every branch of every service that needs kafka in order to pick up an updated Kafka URL. This would mean 1 deploy for kafka plus X amount of deploys for each microservice that needs it.
- If user had `heroku_config` and `heroku_config_association`, it would one deploy just for kafka which would set the configs for each service and let Heroku do the rolling restarts.

## Notes 
- Ideally, I would keep and not deprecate the attributes `config_vars` and `sensitive_config_vars` on `heroku_app`. Users might still want to keep the vars as part of the app resource. These new resources would enable users that have separate repos for their services' terraform configuration to better manage their config variables.


mars:
Glad to see this re-appear after the implementation of app `sensitive_config_vars` 😄👍

A few questions:

-----

```
resource "heroku_config" "foobar-configs" {}
```

How does `heroku_config` store its data? Would it create Heroku app (named `terraform-config-xxxxx`) as its backend? Or do these only store the config var values in the Terraform configuration?

-----

```
    config_vars = "${heroku_config.foobar-configs.vars}"
    sensitive_config_vars = "${heroku_config.foobar-configs.sensitive_vars}"
```

I didn't even realize you can assign whole maps like that 🙃 Can you merge other config vars with this setting?

davidji99:
Hey @mars,

> How does heroku_config store its data? Would it create Heroku app (named terraform-config-xxxxx) as its backend? Or do these only store the config var values in the Terraform configuration?

Currently, I have spec to just store the values in terraform configuration and remote state. I think creating a heroku app with `heroku_config` might be confusing to the end user when the purpose of the resource is to represent a config to be used in heroku? Do you think there would be a lot of value added to creating an app with this new resource?

> I didn't even realize you can assign whole maps like that 🙃 Can you merge other config vars with this setting?

I didn't see Intellij complain when I mocked some sample code to test this theory. I am hoping that it will work in real life. I'll let you know what I get when I work on this PR further.

mars:
> Currently, I have spec to just store the values in terraform configuration and remote state.

👍 Makes sense to have the TF configuration be the central source of these values. I agree that creating a Heroku app to store them would probably be confusing and increase the likelihood of config drift.

> I didn't see Intellij complain when I mocked some sample code to test this theory. I am hoping that it will work in real life. I'll let you know what I get when I work on this PR further.

🤞🤓

davidji99:
@mars:

Confirmed the below works:
```
resource "heroku_config" "config" {
  name = "sample configs"
  vars = {
    VAR1 = "this is string"
  }

  sensitive_vars = {
    DB_URL = "some_url"
  }
}

resource "heroku_config_association" "foobar-config_association" {
  app_id     = "${heroku_app.dont-delete.name}"
  vars = "${heroku_config.config.vars}"
  sensitive_vars = "${heroku_config.config.sensitive_vars}"
}
```

Essentially, I make changes by modifying `heroku_config.config` only and leave  `heroku_config_association` alone.

davidji99:
Deletion behavior is to remove all config vars defined in a `heroku_config_association` block. For example, if one were to remove this entire block:
```
resource "heroku_config_association" "foobar-config_association" {
  app_id     = "${heroku_app.foobar.name}"
  vars = {
    RAKE_ENV = "prod"
    LOG_LEVEL = "debug"
  }

  sensitive_vars = {
    API_TOKEN = "12345678"
    API_KEY = "abcdefg"
  }
}
```
`RAKE_ENV`, `LOG_LEVEL`, `API_TOKEN`, `API_KEY` will be removed from `heroku_app.foobar`. All other existing config vars will remain in place.

davidji99:
## Importing
Import mechanism for `heroku_config` will not be available since there's no remote resource to import from as it is stored in state only.

Import mechanism for `heroku_config_association` might be tricky as it doesn't seem possible to import specific config vars in the `heroku_config_association` block. I rather have the user experience be some complex parsing of a command like this: `terraform import heroku_config_association.foobar <app_id> RAILS_ENV=prod, etc`. However the saving grace here the Platform API does not allow for duplicate config vars. Therefore applying a `heroku_config_association` when the config vars already exist on the app will result in a `noop`.

heroku/resource_heroku_app.go
```diff
@@ -243,19 +243,30 @@ func resourceHerokuAppImport(d *schema.ResourceData, m interface{}) ([]*schema.R
 
 	// Flag organization apps by setting the organization name
 	if app.Organization != nil {
-		d.Set("organization", []map[string]interface{}{
+		setErr := d.Set("organization", []map[string]interface{}{
```
mars: 👏👏👏

OMWow. Nice work handling all these errors.

website/docs/r/config.html.markdown
```diff
@@ -0,0 +1,48 @@
...
+description: |-
+  Provides a Heroku Config resource, making it possible to define variables that can be used in other Heroku terraform resources.
+---
+
+# heroku\_config
+Provides a Heroku Config resource, making it possible to define variables 
+to be used throughout your Heroku terraform configurations. Combined with `heroku_config_association`,
+these two resources enable users to decouple setting config var(s) from the `heroku_app` resource.
+  
+~> **NOTE:** This resource **DOES NOT** make any CRUD operations to your Heroku account.
```
mars: 👍 for this warning. May I suggest a revision to clarify and help guide folks along:

```suggestion
~> **NOTE:** Unlike most Terraform resources, this resource **DOES NOT** by itself create or update anything on Heroku. A [`heroku_config_association`](config_association.html) is required to actually set these values on Heroku apps.
```

davidji99: Thanks, I'll add that.

davidji99:
@mars @talbright @bernerdschaefer review? Thanks.

heroku/resource_heroku_config_association.go
```diff
@@ -0,0 +1,268 @@
...
+		Read:   resourceHerokuConfigAssociationRead,
+		Update: resourceHerokuConfigAssociationUpdate,
+		Delete: resourceHerokuConfigAssociationDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: resourceHerokuConfigAssociationImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"app_id": {
```
davidji99: To avoid (#124), I'd like to start the trend of using `app_id` for child resources. One could still use the app name for this attribute, but I figure the attribute naming would show preference to the app UUID.

dtaylor-sfdc:
Maybe rename `heroku_config_association` to `heroku_app_config_association` and `heroku_config` to `heroku_app_config`? I have no strong opinion on it, just thinking that Heroku might one day add other resources with configs, so may as well future-proof the resource names now, just in case.

davidji99:
This should also handle #47.

davidji99:
> Maybe rename `heroku_config_association` to `heroku_app_config_association` and `heroku_config` to `heroku_app_config`?

I renamed `heroku_config_association` to `heroku_app_config_association` but left `heroku_config` as-is since the latter does not create any app and could be confusing to the end user.

mars:
> Should remove `name` attribute from `heroku_config`?

I don't see how `name` is useful. Since this is Terraform local-state only, the Terraform identifier is all we need.

davidji99:
> > Should remove `name` attribute from `heroku_config`?
> 
> I don't see how `name` is useful. Since this is Terraform local-state only, the Terraform identifier is all we need.

👍 removed.

heroku/resource_heroku_config_association.go
mars: If we know that Heroku API always returns the same value that was set here originally, then making this specific to ID seems fine, but in my experience some resources will return the App name even if you make a request by App ID, which then causes flip-flopping in the state.

davidji99:
> Would it be possible to either make app-specific vars override the shared ones? Or would we want the opposite, shared config vars to override app-specific ones? Or maybe throw a halting error if the same var name is defined in both places?

This is not possible to do since terraform-core doesn't have the capability to have one resource read the configuration of another during a CRUD operation. Having this resource lookup existing config vars that may have been set with `heroku_app.config_vars` wouldn't work because this would disrupt the UPDATE operation. An config var that needs to be updated would already be on the app and there would be no distinction between a var that actually needs to be updated versus an var that was set previously.

davidji99:
@mars, to prevent an infinite loop, I added a function in `heroku_app_config_association` CREATE to get the current app config vars and compare them against the new vars only. It's not possible to do this in UPDATE since the vars have had to already exist for terraform to update them. The error message that is outputted when this scenario happens:

```
	testing.go:531: Step 0, expected error:
		
		Error applying: 1 error occurred:
			* heroku_app_config_association.foobar-config: 1 error occurred:
			* heroku_app_config_association.foobar-config: [ERROR] The following config vars already exist on the app prior to this resource creating them: [RAILS_ENV PRIVATE_KEY]
		To prevent an infinite dirty plan/config drift, please define these vars in terraform in either heroku_app.config_vars OR heroku_app_config_association
```
This should solve this scenario's issue.

mars:
Hi @davidji99 😊

> …to prevent an infinite loop, I added a function…

This sounds reasonable, but when I tried applying a config with conflicting vars, the resulting error awkwardly sticks around.

I begin `terraform apply` with the same vars defined in both `heroku_app.config_vars` & `heroku_app_config_association.vars`. After getting the "config vars already exist on the app prior" error, I try deleting the `config_vars` from app, and also deleting the `heroku_app_config_association`, but it's stuck on that error.

The only configuration solution I found was to completely remove the `heroku_app` resource, so the app's dirty config vars are deleted, and then put it all back.

Do you think it's possible to smooth that edge and make the problem easier to resolve?

> Perhaps we could mark heroku_app.config_vars as DEPRECATED and add a warning about this infinite dirty plan in the docs?

I *would* be up for deprecating them to get all config into this separate associated resource, but the migration path seems pretty rough right now.

I did continue your thread in Hashicorp Slack where `CustomizeDiffFunc` was mentioned as a way to get at other resources state. Hoping to here back that it might offer a sane solution to this 🙈

heroku/resource_heroku_app_config_association_test.go
```diff
@@ -0,0 +1,200 @@
...
+resource "heroku_app" "foobar" {
+    name = "%s"
+    region = "us"
+  organization {
+    name = "%s"
+  }
+}
+
+resource "heroku_app_config_association" "foobar-config" {
+    app_id = "${heroku_app.foobar.name}"
```
mars: Seems like this attribute is named to receive the app's `id` not its `name`.

heroku/resource_heroku_app_config_association_test.go
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    sensitive_vars = {
+        PRIVATE_KEY = "it_is_a_secret"
+        API_TOKEN   = "some_token"
+    }
+}
+
+resource "heroku_app_config_association" "foobar-config" {
+    app_id = "${heroku_app.foobar.name}"
```
mars: Should be the app's `id`.

heroku/resource_heroku_app_config_association_test.go
```diff
@@ -0,0 +1,200 @@
...
```
mars: Should be the app's `id`.

davidji99:
> I begin `terraform apply` with the same vars defined in both `heroku_app.config_vars` & `heroku_app_config_association`. After getting the "config vars already exist on the app prior" error, I try deleting the `config_vars` from app, and also deleting the `heroku_app_config_association`, but it's stuck on that error.
> 
> The only configuration solution I found was to completely remove the `heroku_app` resource, so the app's dirty config vars are deleted, and then put it all back.

Yeah @mars, that was what I was afraid of when I wrote the below:
```
 do however foresee a potential issue if one were to migrate from heroku_app.config_vars to heroku_app_config_association. Essentially, one would need to delete the remove the config var definitions from heroku_app and then add then back immediately after with the new resource.
```

heroku/resource_heroku_config_association.go
davidji99: > Based on how this resource works (purely setting the ID, never getting it), app_id really does seem fine here.

The other reason why I did it this way besides avoiding the issue in #124, `ConfigVarUpdate` doesn't return any information about the parent app. I want to avoid the overhead of making a `GET` just to get the app's information.

mars:
I think this is closer than ever @davidji99 😅

> If I remove this added function…

Removing that check function would result again in infinite apply loops when the same var is defined in an app and in an associated config. So, I don't think that's the right option 👎 

What you mentioned about removing conflicting `config_vars` while leaving the `config_vars` map in-place is the knowledge I was missing!

I've now verified with the current state of this branch that if I define a configuration with the same, conflicting `heroku_app.config_vars` & `heroku_config.vars` that the "config vars already exist on the app prior" error halts apply. Then, I can recover from the error by deleting just the var's entry in the `config_vars` map, leaving `config_vars` in place, even if empty ✅

As long as this minutiae is documented clearly, I think this will be fine for release. Perhaps both the original `heroku_app` and new `heroku_app_config_association` resource should state clearly:

> Deleting a `config_vars` map (or other type of var map) does not actually delete the vars from the Heroku app. To remove an existing variable, leave the `config_vars` map in-place and delete only its entry from the map. If `config_vars` is empty, then only after applying the change, it may be deleted from the configuration.

davidji99:
@mars,

Screen Shot 2019-03-27 at 10 05 44

^ For this PR, this behavior is only applicable to `heroku_app.config_vars` & `heroku_app.sensitive_config_vars` as it is the current behavior for those attributes. I don't think this is the correct behavior so I 'fixed' it for `heroku_app_config_association.vars` & `heroku_app_config_association.sensitive_vars`. The behavior for these attributes on ` `heroku_app_config_association` is if one removes the entire attribute(s), the resource will delete any config vars defined in them on the app remotely. 

Therefore, I can add what you suggested in the screenshot above to `heroku_app` and maybe an additional blurb on `heroku_app_config_association` on how to best migrate to using this new resource from 1) `heroku_app` or 2) manually

davidji99:
@mars review. Should be good to go now 🤞

mars:
👀

website/docs/r/config.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+
+## Argument Reference
+* `vars` - Map of vars that are can be outputted in plaintext
+* `sensitive_vars` - This is the same as `vars`. The main difference between the two
+attributes is `sensitive_vars` outputs are redacted on-screen and replaced by a <sensitive> placeholder, following a terraform
+plan or apply. It is recommended to put private keys, passwords, etc in this argument.
+
+## Attributes Reference
+The following attributes are exported:
+* `id` - The ID of the config
```
mars: Screen Shot 2019-03-29 at 11 22 10


A linebreak is required here to format correctly:

```suggestion

* `id` - The ID of the config
```

website/docs/r/app.html.markdown
```diff
@@ -80,10 +80,14 @@ The `organization` block supports:
 * `locked` (boolean)
 * `personal` (boolean)
 
-~> **NOTE:** Removing an entire `config_vars` or `sensitive_config_vars` block from a
-configuration will not actually remove the vars on the remote resource.
-This is especially important if you are migrating all `config_vars` to `sensitive_config_vars`.
-Leave the empty `config_vars` or `sensitive_config_vars` block in place to remove previously defined vars.
+~> **NOTE:**
+1. Deleting an entire `config_vars` or `sensitive_config_vars` map from a `heroku_app`
```
mars: Screen Shot 2019-03-29 at 11 24 47


Can we relate this information more clearly?

I suggest this becomes a normal text section with a header:

```
### Deleting vars

Deleting an entire `config_vars` or `sensitive_config_vars` map from a `heroku_app`
configuration will not actually remove the vars on the remote resource. To remove an existing variable,
leave these attribute maps in-place and delete only its entries from the map. Once these attributes are
empty, the map itself may be deleted from the configuration. Otherwise, if one deletes the map with existing
entries, the config vars will not be deleted from the remote resource.

This is especially important if you are migrating all `config_vars` to `sensitive_config_vars` or migrating
config vars to `heroku_app_config_association` resource.
```

And then link to that header's auto-generated anchor:

```
* `config_vars`<sup>[1](#deleting-vars)</sup> - (Optional) Configuration variables for the application…
```

website/docs/r/app_config_association.html.markdown
```diff
@@ -0,0 +1,82 @@
...
+
+* User has separate git repositories for various micro-services. Multiple micro-services use Kafka.
+* User has a separate repository for kafka terraform files with blue/green support.
+* User builds out new clusters.
+* Prior to this resource's introduction, user would need one `terraform apply` to update state and X number of `terraform apply`
+for each micro-service to pick up the new kafka clusters. However with this resource, user can do one `terraform apply`
+and let Heroku handle the rolling restarts to pick up the new config vars.
+
+~> **NOTES:**
+* Heroku does not have a 'sensitivity' distinction for its config variables.
```
mars: Screen Shot 2019-03-29 at 11 38 45

Because lists inside Notes don't render as lists, I suggest making these two normal text sections:

```
### "Sensitive" is not secret

Heroku does not have a 'sensitivity' distinction for its config variables. This distinction is only made during Terraform `plan` and `apply` to avoid leaking sensitive data in the console output.

### Beware conflicting vars

Be careful when having config variables defined in both `heroku_app` and `heroku_app_config_association` resources. As the latter resource
has a dependency on the former, any overlapping config variables in `heroku_app` will be overwritten in `heroku_app_config_association`.
Furthermore, this overlap will cause an infinite dirty 'terraform plan' if config variables have different values on both resources
at the same time. It is recommended to use one or the other resource, not both, to manage your app(s) config vars.
```

website/docs/r/app_config_association.html.markdown
```diff
@@ -0,0 +1,82 @@
...
+## Argument Reference
+* `app_id` - (Required) A Heroku app's `UUID`. Can also be the name of the Heroku app but `UUID` is preferred as it is idempotent
+* `vars` - Map of config vars that are can be outputted in plaintext
+* `sensitive_vars` - This is the same as `vars`. The main difference between the two
+attributes is `sensitive_vars` outputs are redacted on-screen and replaced by a <sensitive> placeholder, following a terraform
+plan or apply. It is recommended to put private keys, passwords, etc in this argument.
+
+## Attributes Reference
+The following attributes are exported:
+* `id` - The ID of the app config association
```
mars: Screen Shot 2019-03-29 at 11 42 21

A linebreak should fix this:

```suggestion

* `id` - The ID of the app config association
```

davidji99:
will take a look @mars thanks! Forgot about the `make website` command, which I think you used to generate the site?

mars:
Yes, `make website` (with Docker running) is the command. And navigate directly to the Heroku provider path (some intermediate pages throw errors).

davidji99:
Done @mars

talbright:
lots of hard work went into this one...congrats 🎆

PR #184, opened 2019-03-19
Create a resource for creating Heroku team invitation

### LINKED ISSUE #68: Create a resource for creating Heroku teams
https://devcenter.heroku.com/articles/platform-api-reference#team-create

davidji99:
## EDIT
It seems once a team invitation has been accepted, the invitation itself is deleted. This would be problematic for the provider after every state refresh. Therefore, I wonder about the usability of a `heroku_team_invitation` resource.

## SCOPE
New resource: `heroku_team_invitation`:
```
resource "heroku_team" "foobar" {
    team_id = "my_team_id"
    email = "userof@team.com"
    role = "member"
}
```

Reference: https://devcenter.heroku.com/articles/platform-api-reference#team-invitation

TODO:
- [ ] Create
- [ ] Update
- [ ] Read
- [ ] Delete
- [ ] Import
- [ ] Tests
- [ ] Docs


mars:
This functionality requires an Enterprise Account, and that the Terraform user must be an Admin with team creation permissions. I see there’s a notice in the new resource docs, but I think we need to refine to match the Account/Team terminology used by Sales & Support.

Standard Teams have a credit card signup flow that cannot be automated.

Also, the `credit_card_collections` & `provisioned_licenses` fields (while documented in the Platform API reference) seems like internal fields that may reflect some state of the account but are not designed to be set by Heroku users.

I’ll investigate with Heroku product folks.

davidji99:
> This functionality requires an Enterprise Account, and that the Terraform user must be an Admin with team creation permissions. I see there’s a notice in the new resource docs, but I think we need to refine to match the Account/Team terminology used by Sales & Support.

Are you sure @mars? I was able to create a non-enterprise team with the proposed `heroku_team` resource. The API docs say the `POST` only requires `name`.

I think I was hoping to add this resource so people can import their enterprise teams and then a new `heroku_team_invitation` to manage access (https://devcenter.heroku.com/articles/platform-api-reference#team-invitation).

mars:
I have some news from my Heroku-internal research:

`credit_card_collections`, `provisioned_licenses`, `type`, & `membership_limit` are all read-only fields that will ignore updates.

There's currently no generally available Heroku API that will allow you to explicitly associate a new team with an Enterprise Account. Currently, if the authorized user is a member of multiple Enterprise Accounts, it's not clear what the logic is for "where" teams are created.

@davidji99 the fact that you can create a Team is a result of being in an Enterprise Account.

davidji99:
> @davidji99 the fact that you can create a Team with only a `name` is a result of being in an Enterprise Account. Otherwise, credit card info is required.

Thanks for the info. Perhaps maybe the Platform Docs need to be updated with this exception as in its current form, one could create a regular team with just `name`.

davidji99:
The path moving forward for this PR is to create a data resource instead for `heroku_team`.

PR #188, opened 2019-03-27
New data resource: Heroku Team

davidji99:
New data source for heroku teams and heroku enterprise teams:

```
data "heroku_team" "my_heroku_team" {
  name = "name_of_my_heroku_team"
}
```

Sample output in state:
```
                "data.heroku_team.my_heroku_team": {
                    "type": "heroku_team",
                    "depends_on": [],
                    "primary": {
                        "id": "<some_uuid>",
                        "attributes": {
                            "default": "false",
                            "id": "<some_uuid>",
                            "membership_limit": "500",
                            "name": "<some_team>",
                            "provisioned_licenses": "true"
                        },
                        "meta": {},
                        "tainted": false
                    },
                    "deposed": [],
                    "provider": "provider.heroku"
                },
```

### DIFF
--- a/helper/test/config.go
+++ b/helper/test/config.go
@@ -19,6 +19,7 @@ const (
 	TestConfigSpaceOrganizationKey
 	TestConfigSlugIDKey
 	TestConfigEmail
+	TestConfigTeam
 )
 
 var testConfigKeyToEnvName = map[TestConfigKey]string{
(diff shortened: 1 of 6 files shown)


davidji99:
@mars @talbright review?

davidji99:
> Hi @davidji99! Looks good. One question though: what about `role` and `type`? Those seem like useful pieces of info (https://devcenter.heroku.com/articles/platform-api-reference#team)

Hi @talbright, 

From my understanding of `role`, I think this attribute's value could change depending on the user that makes the API request to `/team/{team_id}`. So I didn't see a value to having them on the data source as they aren't really information about the team itself. Hover, I think `type` is good to have so I'll add that in to the data source. Thanks!

PR #195, opened 2019-05-03
Upgrade to Go Modules-compatible Heroku client

mars:
Validate the [heroku-go **Go 1.11 module support** PR](https://github.com/heroku/heroku-go/pull/39), and then upgrade to it once released. At that time, this PR will become mergeable.


mars:
[Hashicorp CI was successful for all but one](https://ci-oss.hashicorp.engineering/viewLog.html?buildTypeId=HerokuProvider_ProviderHeroku&buildId=55406&branch_HerokuProvider_ProviderHeroku=195) flappy `heroku_space_vpn_connection` test, so I'm calling it good ✅

go.mod
```diff
@@ -20,7 +20,7 @@ require (
 	github.com/hashicorp/hcl2 v0.0.0-20180714220537-966851f3097f // indirect
 	github.com/hashicorp/terraform v0.11.12-beta1.0.20190214175014-182daa619826
 	github.com/hashicorp/yamux v0.0.0-20180604194846-3520598351bb // indirect
-	github.com/heroku/heroku-go v0.0.0-20190103224148-ad17585a922f
+	github.com/heroku/heroku-go v4.0.0+incompatible
```
talbright: Wasn't expecting this, taking a look at https://github.com/golang/go/wiki/Modules#can-a-module-consume-a-v2-package-that-has-not-opted-into-modules-what-does-incompatible-mean

talbright:
Did you follow these steps: https://github.com/terraform-providers/terraform-provider-heroku#updating-or-adding-dependencies

go.mod
mars: It makes sense, because that `v4.0.0` tag is incompatible with the previous version that was installed, but it does seem weird and broken when installed fresh too.

Here's `go.mod` just created for the simple [heroku-go example](https://github.com/heroku/heroku-go#example):

```
module github.com/mars/heroku-go-fresh

go 1.12

require (
	github.com/cenkalti/backoff v2.1.1+incompatible // indirect
	github.com/google/go-querystring v1.0.0 // indirect
	github.com/heroku/heroku-go v4.0.0+incompatible // indirect
	github.com/pborman/uuid v1.2.0 // indirect
)
```

But it seems that the `+incompatible` is going to hang out with us from now one since we "broke" the Go versioning expectations.

mars:
> Did you follow these steps…

Yes, exactly.

go.mod
mars: Note the `backoff` module has suffered the same `+incompatible` fate 😂

Do we care? Seems like Go is making this look way worse than it is.

talbright: That's what I'm wondering, if I'm overthinking this or not 🙄

bernerdschaefer:
I unfortunately don't have much/any experience with go modules (as a user or package author), so I can't really weigh in meaningfully on this.

go.mod
talbright: Well now I'm thinking I made a mistake and should have kept the `vN` notation in heroku-go 

As a result of Semantic Import Versioning, _code opting in to Go modules must comply with these rules_:
* Follow semver.
* If the module is version v2 or higher, the major version of the module must be included as a /vN at the end of the module paths used in go.mod files (e.g., module github.com/my/mod/v2, require github.com/my/mod/v2 v2.0.0) and in the package import path (e.g., import "github.com/my/mod/v2/mypkg").
* If the module is version v0 or v1, do not include the major version in either the module path or the import path.

In other words I did all the right things in the heroku-go PR, except that I should have had v4 in the package. My bad.

mars: Don’t sweat it Trent 😌

I think the right thing to do will be: open a new PR on heroku-go putting a `/v5` path suffix in place, and then tagging the release `v5.0.0`.

talbright: Thanks 😊  I'll take care of it.

mars:
Closing in favor of the newer #196

PR #198, opened 2019-05-22
Fix minor documentation inaccuracy

pgjones:
Without this fix terraform will output this error

    Error reading depends_on for heroku_cert[...]: At ... unknown slice type: *ast.LiteralType


mars:
Thank you @pgjones 😄

PR #204, opened 2019-06-09
Correct example to use assignment

glenngillen:
Setting `config_vars` in this context is via assignment and not a block.


davidji99:
looks good!

Issue #209, opened 2019-06-26
Support a means of attaching the same add-ons of one app to another


robertcnix:
This is a feature request.

It would be incredibly useful if the `heroku_app` data resource returned the attached add-ons. Or maybe even more useful, the `heroku_addon` data resource could accept an `app` name in addition to the add-on name (only one _required_).

The value here is with respect to data-driven infrastructure. Being able to define a list of apps that contain the add-ons needed for another app allows for the separation of infrastructure `.tf` definitions from the description of a service-app (in `.tfvars`). For example, I'd like developers to be able to specify the add-ons their service needs in a simple list in a `.tfvars` file without having to touch any `.tf` files.

The problem is that the only place the add-on name is provided seems to be the output of the `heroku_addon` resource so if the definition of the add-on is in another terraform definition, it requires accessing the state of the other resource which is more cumbersome than using data resources.

Additional, small gripe, this is a little misleading, implying that the `name` is an app name but it's actually the add-on name:

```
data "heroku_addon" "from_another_app" {
  name = "addon-from-another-app"
}
```

### Terraform Version
```
Terraform v0.12.3
+ provider.heroku v2.0.1
```

### Affected Resource(s)
*data*  resources:
- `heroku_app`
- `heroku_addon`

### Terraform Configuration Files
It would be nice to do something like this:
```hcl
data "heroku_addon" "addons" {
	app = "addons-from-this-app"
}

resource "heroku_addon_attachment" "attachments" {
	app_id  = "${heroku_app.default.id}"
	addon_ids = "${data.heroku_addon.addons}"
}
```

where `data.heroku_addon.addons` is a list and allows for attaching all the same add-ons to another app.

bernerdschaefer:
I'm not certain about fetching all addons or attaching many addons at once, but the platform API does support setting custom names on addons as well as [fetching an addon by that custom name, namespaced by app](https://devcenter.heroku.com/articles/platform-api-reference#add-on-info-by-app).

The provider already supports [giving add-ons a meaningful name](https://www.terraform.io/docs/providers/heroku/r/addon.html#name) (see [DevCenter on meaningful names](https://devcenter.heroku.com/articles/add-ons#meaningful-names)):

```tf
resource "heroku_addon" "messaging" {
  app = "myapp"
  plan = "heroku-kafka:private-5"
  name = "messaging"
}
```

So if the data resource supported providing an app name / id, we could look up the add-on without knowing its globally unique name:

```tf
[...]

PR #210, opened 2019-06-27
Ability to set addon name

davidji99:
- Need to update `heroku-go` to support passing in `Name` during a `PATCH` addon request.

Related to #209.

### DIFF
--- a/heroku/resource_heroku_addon.go
+++ b/heroku/resource_heroku_addon.go
@@ -3,6 +3,7 @@ package heroku
 import (
 	"context"
 	"fmt"
+	"github.com/hashicorp/terraform/helper/validation"
 	"log"
 	"net/url"
 	"regexp"
@@ -13,10 +14,6 @@ import (
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	heroku "github.com/heroku/heroku-go/v5"
-)
-
-const (
-	AddonNameMaxLength = 256
 )
 
 // Global lock to prevent parallelism for heroku_addon since
@@ -52,9 +49,10 @@ func resourceHerokuAddon() *schema.Resource {
 			},
 
 			"name": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validateCustomAddonName,
 			},
 
 			"config": {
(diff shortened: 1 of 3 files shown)


heroku/resource_heroku_addon.go
```diff
@@ -233,3 +256,32 @@ func AddOnStateRefreshFunc(client *heroku.Service, appID, addOnID string) resour
...
+
+// validateAddonName uses the documented regex expression to make sure the user provided addon name is valid.
+//
+// Reference: https://devcenter.heroku.com/articles/platform-api-reference#add-on-create-optional-parameters
+func validateAddonName(name string) error {
+	errors := make([]string, 0)
+
+	// First validate length. There is no documented length
+	// but I've tried up to 256 characters so this will be max for now.
+	if len(name) > AddonNameMaxLength {
```
bernerdschaefer: I think this could be moved into a `ValidateFunc` field on the attribute, and use the [validation.StrLenBetween](https://godoc.org/github.com/hashicorp/terraform/helper/validation#StringLenBetween) helper.

heroku/resource_heroku_addon.go
```diff
@@ -233,3 +256,32 @@ func AddOnStateRefreshFunc(client *heroku.Service, appID, addOnID string) resour
...
+		errors = append(errors, fmt.Sprintf("Length cannot exceed %v characters", AddonNameMaxLength))
+	}
+
+	// Then validate the content of the string against documented regex.
+	regex := regexp.MustCompile(`^[a-zA-Z][A-Za-z0-9_-]+$`)
+	matches := regex.FindStringSubmatch(name)
```
bernerdschaefer: This could also be moved into an attribute-level validator using the [validation.StringMatch](https://godoc.org/github.com/hashicorp/terraform/helper/validation#StringMatch) helper.

heroku/resource_heroku_addon.go
```diff
@@ -152,20 +168,27 @@ func resourceHerokuAddonRead(d *schema.ResourceData, meta interface{}) error {
...
-			context.TODO(), app, d.Id(), heroku.AddOnUpdateOpts{Plan: d.Get("plan").(string)})
-		if err != nil {
-			return err
-		}
+		opts.Plan = d.Get("plan").(string)
+	}
 
-		// Store the new ID
-		d.SetId(ad.ID)
+	// TODO: uncomment once the go client supports this
```
bernerdschaefer: We could also set `ForceNew: true` on the attribute, and note in the docs that renaming an add-on is not currently supported, rather than block on updating the schema.

heroku/resource_heroku_addon.go
davidji99: Ah right...forgot about the `ValidateFunc`.

heroku/resource_heroku_addon.go
davidji99: 👍

heroku/resource_heroku_addon.go
davidji99: Ah by now (via slack msgs), this attribute is indeed updatable. So the provider should support this attribute update as well.

heroku/resource_heroku_addon.go
davidji99: Yeah @bernerdschaefer, using the `validation.*` packages might not work since we need to use two of them. Looking at the AWS provider, it seems one either uses one of the default validations or writes their own if they need more than one.

davidji99:
@bernerdschaefer reworked the validation.

davidji99:
Once #211 is in, I will be able to continue working on this.

mars:
#211 merged and [released in v2.0.2](https://github.com/terraform-providers/terraform-provider-heroku/releases/tag/v2.0.2)

davidji99:
This is ready to go.

mars:
Hashicorp CI passed for this branch ✅

Mergin' 🤠

PR #212, opened 2019-07-01
Change config from block to attribute assignment

naliwajek:
Terraform 0.12.x requires config to be an attribute for me, not a block.


bernerdschaefer:
Thanks!

PR #218, opened 2019-07-09
Add support for reading addon config vars

### LINKED ISSUE #140: Expose heroku_addon config_vars values
I need to extract the value of a configuration variable that a `heroku_addon` adds to my `heroku_app`. This is to configure an AWS lambda function that needs to connect to this addon.

However, the `config_vars` attribute on `heroku_addon` does not export a map of configuration variables and their values, but only a list of the variables' names.

Here's what a typical config may look like:

```
resource "heroku_app" "worker" {
  app  = "worker"
}

resource "heroku_addon" "cloudamqp" {
  app  = "${heroku_app.worker.name}"
  plan = "cloudamqp:lemur"
}

output "cloudamqp_config_vars" {
  value = "${heroku_addon.cloudamqp.config_vars}"
}
```

And the output it gives.

```
cloudamqp_config_vars = [
    CLOUDAMQP_APIKEY,
    CLOUDAMQP_URL
]
```

Would it be possible to get a map instead, so I could extract the value I need?

Photonios:
Fixes #140 

This is my first attempt. Please do comment. I'd also need some suggestions on how to properly test this.

- --

There's also the case of this change breaking backwards compatibility. I've modified the type of `heroku_addon.config_vars` from `schema.TypeList` to `schema.TypeMap`. 

If there are any suggestions or ideas on how to make the change without making a breaking change. I am all ears.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+## 2.6.0 (Unreleased)
 ## 2.5.0 (July 03, 2020)
 
 FEATURES:


heroku/data_source_heroku_addon.go
```diff
@@ -34,11 +34,8 @@ func dataSourceHerokuAddon() *schema.Resource {
 			},
 
 			"config_vars": {
-				Type:     schema.TypeList,
```
davidji99: The original attribute `Type` definition for `"config_vars"` is correct. The `addon` resource returns a `config_vars` list in the API call. For example, a `GET` on the heroku postgres addon would return:
```
    "config_vars": [
        "DATABASE_URL"
    ],
```

Please roll back your proposed change.

heroku/resource_heroku_addon.go
```diff
@@ -63,11 +63,8 @@ func resourceHerokuAddon() *schema.Resource {
 			},
 
 			"config_vars": {
-				Type:     schema.TypeList,
```
davidji99: Same comment as https://github.com/terraform-providers/terraform-provider-heroku/pull/218/files#r301860291

davidji99:
Hi @Photonios,

Thanks for your PR! However, there is an issue with your proposed change.

Echoing my comment I made on [L37](https://github.com/terraform-providers/terraform-provider-heroku/pull/218/files#diff-eea0541982e3d6478f07491aee62e4dcL37), you are confusing `heroku_addon.config_vars` attribute with an [addon's configuration](https://devcenter.heroku.com/articles/platform-api-reference#add-on-config).

`heroku_addon.config_vars` represents this [(reference](https://devcenter.heroku.com/articles/platform-api-reference#add-on)):
Screen Shot 2019-07-10 at 11 09 54

So the current implementation of `heroku_addon.config_vars` is correct. Your PR's attempt to change this attribute's `Type` as well as what data it represents is a breaking change to the provider and not in line with what is returned by the Heroku Platform API. If you wish to expose the addon's configuration as a data source, I think a separate `data_source_heroku_addon_config` could be more appropriate here since how the configuration is a child resource of the addon. This way, the original `data_source_heroku_addon` remains intact. 

If we were to expose the addon's configuration to `data_source_heroku_addon`, you could create a new attribute called `config` or `configuration` to represent the addon configuration. However `data_source_heroku_addon.config` would bit a confusing to `resource_heroku_addon.config` because they are different things. For the addon resource, the `config` attribute represents:
Screen Shot 2019-07-10 at 11 22 25

The last thing with trying to expose an addon's configuration is the configurations are set asynchronously so we would need a polling mechanism. Otherwise, users might need to refresh their statefiles after X amount of time.

So as you can see, there needs to be a bit of alignment in naming convention with addons if we are to expose the configuration values. I'm thinking a separate data source would be a simple path forward as a start.

Photonios:
Hi @davidji99 !

Thanks for the quick turnaround. Much appreciated. 

I am aware that this would've been a breaking change, not in line with Heroku's Platform API. I was doubting between this and adding a new data source. Since this was cheaper to implement, I initially went with this approach to test the waters so to say. Obviously, an extra data source is cleaner.

I believe the separate data source (`data_heroku_addon_config`) is more appropriate as well since it's more in line with how the Heroku Platform API exposes things.

As for the polling, I am aware they are set async. I think we can safely assume it's ready and set when the addon is in the `provisioned` state. This would mean we need a similar approach as `resource_heroku_addon` takes, waiting for the addon to be in the `provisioned` state. The alternative would be to wait till they are set. But I believe the former to be more reliable. If you have another opinion, please let me know.

Thanks for the pointers on the tests and docs. I'll make sure to that in the next iteration.

I'll get to work :-) Thanks again for taking the time!

davidji99:
No problem! 

A few responses:

> I am aware that this would've been a breaking change, not in line with Heroku's Platform API.

Gotcha. I originally read your "breaking backwards compatibility" statement as a provider attribute type breaking change.

----

> As for the polling, I am aware they are set async. I think we can safely assume it's ready and set when the addon is in the `provisioned` state.

Yeah, I've chatted with the Heroku engineers about this in the past. Unless the mechanism/process has changed since then, we can't safely assume the configs are ready and set when the `addon` is provisioned =(. They have mentioned some addon providers can take longer than 5 minutes to set their addon's configuration. This is one of the reasons why exposing the addon configuration data hasn't been implemented in the provider yet because the discussions back then couldn't come to a consensus on the ideal experience around polling for those configs.

---- 
> This would mean we need a similar approach as `resource_heroku_addon` takes, waiting for the addon to be in the `provisioned` state.

Since the potentially new data source of `data_heroku_addon_config` will rely on `resource_heroku_addon`, I believe `terraform` will create a dependency so the data source won't get refreshed until the resource is ready, at least on initial resource creation. For example, I foresee the code to be:
```
data "heroku_addon_config" "foobar-config" {
  addon_id = heroku_addon.foobar.id
}
```

----

This provider has the ability to set custom attributes at the `provider {}` block level. We can probably leverage this to set a generic timeout for polling for addon configuration that the user can further set themselves. Maybe for the polling mechanism itself, since we can get the config vars, we can poll until all the vars have values?

Photonios:
> Yeah, I've chatted with the Heroku engineers about this in the past. Unless the mechanism/process has changed since then, we can't safely assume the configs are ready and set when the addon is provisioned =(. They have mentioned some addon providers can take longer than 5 minutes to set their addon's configuration. This is one of the reasons why exposing the addon configuration data hasn't been implemented in the provider yet because the discussions back then couldn't come to a consensus on the ideal experience around polling for those configs.

That's unfortunate :) 

I don't think we can rely on the `heroku_addon_config` data source to always be dependent on a `heroku_addon` resource. You might want to refer to an existing addon that is not being managed by Terraform and it might be that it's being provisioned.

---

> This provider has the ability to set custom attributes at the provider {} block level. We can probably leverage this to set a generic timeout for polling for addon configuration that the user can further set themselves.

As for the timeout, currently, the `heroku_addon` resouce waits up to 20 minutes for an addon to be "provisioned". We can either add a configuration option on a provider level, or simply use the same timeout. How about adding:

* `resource_ready_timeout` - Maximum amount of seconds to wait for a resource to be ready.

20 minutes is a bit long. Maybe we can set the default to 10 and use that for all resources that need do some form of polling?

---

> Maybe for the polling mechanism itself, since we can get the config vars, we can poll until all the vars have values?

As far as I know, the addon provider doesn't have to specify upfront which config vars it is going to set, we can't implement a generic mechanism to wait for all config vars to be set since we don't know which config vars are going to be set by the addon provider. I'd like to propose to add an option to the `heroku_addon_config` data source:

* `config_vars` - List of config var names to retrieve

The computed `config` property would give a map with the config var values specified in `config_vars`.  Config vars that are not in the list are not waited for. They might be returned, but the data source would not wait for them to become available. For clarity's sake, we could only return the whitelisted config vars.

````
data "heroku_addon_config" "bla" {
      name = "myaddon-123"
      config_vars = ["URL", "ID"]
}
```

I think it would be good for this data source to take a `name` option, exactly like the `heroku_addon` data source does. Instead of providing an ID. The ID's are usually not exposed to the user from the Heroku dashboard.

mars:
>> This would mean we need a similar approach as resource_heroku_addon takes, waiting for the addon to be in the provisioned state.
>
> Since the potentially new data source of data_heroku_addon_config will rely on resource_heroku_addon, I believe terraform will create a dependency so the data source won't get refreshed until the resource is ready, at least on initial resource creation. For example, I foresee the code to be…

Terraform data sources can cause state churn (apply/plan will always show diffs) when used with dependencies. See [Data Resource Dependencies](https://www.terraform.io/docs/configuration/data-sources.html#data-resource-dependencies). Having a Terraform configuration that never applies cleanly can be a real pain for automation.

> Maybe for the polling mechanism itself, since we can get the config vars, we can poll until all the vars have values?

Unfortunately there's no generic way to know what Add-on config vars should be set for any given add-on, and add-on providers are free to arbitrarily update them (such as during provisioning and for credential rotation).

@Photonios instead of trying to rig this into our Terraform Provider, which as mentioned has been attempted before and not likely to work any better now, you might be better off writing one-off [External data sources](https://www.terraform.io/docs/providers/external/data_source.html) for each add-on type you want to query.

mars:
Sorry this thread is so long, I just realized you're suggesting that the data source have a filter for specific, explicit add-on config var retrieval. That might work! 👌

Photonios:
My apologies if this comes across as an attempt to hack in a quick fix that works for me but not others. Not my intention. We already solved this internally with an external data source. This little contribution is our attempt and doing something nicer that benefits us and the wider community. Fully committed to finding a solution that's clear and works well for everybody. 

As you saw, I proposed to have a filter for explicit addon config retrieval. I think this would work. I am currently working on implementing what I proposed so we can take it from there. Please bear with me as this is not a full time job for me :-)

Thanks for jumping in. The more eyes the better :-)

davidji99:
> Unfortunately there's no generic way to know what Add-on config vars should be set for any given add-on, and add-on providers are free to arbitrarily update them (such as during provisioning and for credential rotation).

I thought the data resource could query the existing addon to get its `config_vars` and then use that to see if those vars have actually been set? Or are you saying, addon `config_vars` aren't necessarily known even after provisioning?

> My apologies if this comes across as an attempt to hack in a quick fix that works for me but not others. 

No worries at all! This has been in discussion for a bit so when you pulled up a PR, it brings the implementation into focus in order to make it work for everyone well.

> Please bear with me as this is not a full time job for me :-)

All good. Most of the contributors make changes, edits, and reviews in our spare time as well. ^_^

mars:
> I thought the data resource could query the existing addon to get its config_vars and then use that to see if those vars have actually been set? Or are you saying, addon config_vars aren't necessarily known even after provisioning?

It's not clear to me the timing behavior of [Listing Add-on config vars](https://devcenter.heroku.com/articles/platform-api-reference#add-on-config-list). Are the variables returned always the same ones as are already set on the app? Are all possible variables listed but empty until set, or do they only appear when set?

I'm checking with folks internally to hopefully get some clarity. Will follow-up.

mars:
📰 word is that **Add-on config vars only appear once they're set by the provider**, so that cannot be used to know what to wait for.

The proposed data source filter `config_vars = ["URL", "ID"]` seems like the best way forward.

Please, make the `config_var` attribute return immediately though, with empty values for the unset vars.

Then, implement another attribute `await_config_vars = ["URL", "ID"]` that will [retry](https://www.terraform.io/docs/extend/resources.html#retry-helpers), only returning when all the vars appear.

Thanks @Photonios for your work on this! It will be a very useful feature!

mars:
Actually, it looks like the [Add-on](https://devcenter.heroku.com/articles/platform-api-reference#add-on) resource returns state of the add-on, e.g. `provisioning` or `provisioned`.

That endpoint could be polled until `provisioned`, and then the Add-on config vars will all be there!

mars:
FYI `heroku_addon` already [waits for the add-on to be provisioned](https://github.com/terraform-providers/terraform-provider-heroku/blob/master/heroku/resource_heroku_addon.go#L105), so that should be pretty simple to use for this new data source.

PR #221, opened 2019-07-12
Grammar fix

aedades:
Minor grammar fix, choosing one over the other between "are" and "can be".


website/docs/r/app_config_association.html.markdown
```diff
@@ -71,7 +71,7 @@ resource "heroku_app_config_association" "foobar2" {
 ## Argument Reference
 
 * `app_id` - (Required) A Heroku app's `UUID`. Can also be the name of the Heroku app but `UUID` is preferred as it is idempotent.
-* `vars` - Map of config vars that are can be outputted in plaintext.
+* `vars` - Map of config vars that are output in plaintext.
```
davidji99: "Output" should be in past tense here no?

mars: "Output" is multi-tense: present, past, & future all in a single word 👌

PR #231, opened 2019-08-06
Allow specifying initial attachment name when creating addon.

### LINKED ISSUE #227: Allow specifying initial attachment name when creating addon.
There appears to be no way of specifying the initial attachment name when creating an addon (--as parameter on the Heroku command line, also appears to be referred to in the docs as an alias).

This is exposed fairly nicely in the Heroku Go client, https://github.com/heroku/heroku-go/blob/daded33ff8cdcd8127dcfc3c744686ae2e8d4efc/v5/heroku.go#L388-L390

It doesn't seem that far off in it's current state, as it is constructing this very same object, but providing only Name, Config and Plan (everything but Attachment!): https://github.com/terraform-providers/terraform-provider-heroku/blob/4f59dde/heroku/resource_heroku_addon.go#L99

davidji99:
Resolves #227

### DIFF
--- a/heroku/resource_heroku_addon.go
+++ b/heroku/resource_heroku_addon.go
@@ -55,6 +55,12 @@ func resourceHerokuAddon() *schema.Resource {
 				ValidateFunc: validateCustomAddonName,
 			},
 
+			"attachment_name": {
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Optional: true,
+			},
+
 			"config": {
 				Type:     schema.TypeMap,
 				Optional: true,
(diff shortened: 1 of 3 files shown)


davidji99:
From what I experience, the Heroku Platform API does two things:
- Does not return `attachment.name` in the either the `POST` or `GET` call to `/apps/{app_id_or_name}/addons`. It does show up in the UI:
Screen Shot 2019-08-06 at 3 22 24 PM

- Does not allow you to update the attachment name as the following error is returned when attempting to do so:

```
{
    "id": "invalid_params",
    "message": "Unknown params: attachment:name."
}
```

Therefore, the new `heroku_addon.attachment_name` attribute will be set to `ForceNew` as it cannot be updated, only set on initial creation.

----
Questions to answer:
- [ ] What happens when someone has a `heroku_addon_attachment` resource that specifies a different name?

PR #238, opened 2019-09-11
Upgrade TF to v0.12.8

appilon:
Updates TF SDK to v0.12.8 and deletes commented out provider setup test


heroku/provider_test.go
```diff
@@ -68,32 +68,28 @@ func TestProviderConfigureUsesHeadersForClient(t *testing.T) {
...
-//	// Create a dummy netrc file
-//	tmpfileNetrc, err := createTempConfigFile(`machine api.heroku.com login email_login password api_key`, ".netrc")
-//	if err != nil {
-//		t.Fatal(err.Error())
-//	}
-//	defer os.Remove(tmpfileNetrc.Name())
-//	os.Setenv("NETRC_PATH", tmpfileNetrc.Name())
-//	defer os.Unsetenv("NETRC_PATH")
-//	raw := make(map[string]interface{})
-//	rawConfig, err := config.NewRawConfig(raw)
```
mars: Is this the interface that changed? It seems like it's only used in this commented out code, which should probably be deleted.

appilon: Yes this is the only change i was targeting. If this can be deleted, i am happy to do that and relabel this PR as a dependency update.

mars: Yes, let’s please delete the commented code, and then this PR will really be super good to go 😄

appilon:
CI passed!

PR #239, opened 2019-09-13
Add ability to manage App Webhooks

CGA1123:
Adds the ability to manage [Heroku App Webhooks](https://devcenter.heroku.com/categories/app-webhooks) as terraform resources.

There are still some less basic test cases to cover, however we would appreciate reviews nonetheless 😃.

It is worth mentioning that the management of the `secret` + `authorization` is a bit awkward, as there is no way to unset them.

### DIFF
--- a/heroku/provider.go
+++ b/heroku/provider.go
@@ -71,6 +71,7 @@ func Provider() terraform.ResourceProvider {
 			"heroku_app_config_association":            resourceHerokuAppConfigAssociation(),
 			"heroku_app_feature":                       resourceHerokuAppFeature(),
 			"heroku_app_release":                       resourceHerokuAppRelease(),
+			"heroku_app_webhook":                       resourceHerokuAppWebhook(),
 			"heroku_build":                             resourceHerokuBuild(),
 			"heroku_cert":                              resourceHerokuCert(),
 			"heroku_config":                            resourceHerokuConfig(),
--- a/website/docs/r/app_webhook.html.markdown
+++ b/website/docs/r/app_webhook.html.markdown
@@ -1,0 +1,47 @@
+---
+layout: "heroku"
+page_title: "Heroku: heroku_app_webhook"
+sidebar_current: "docs-heroku-resource-app-webhook"
+description: |-
+  Provides the ability to manage application webhooks
+---
+
+# heroku\_app\_webhook
+
+Provides a [Heroku App Webhook](https://devcenter.heroku.com/categories/app-webhooks).
+
+## Example Usage
+
+```hcl
+# Create a new Heroku app
+resource "heroku_app" "foobar" {
+  name = "foobar"
+}
+
+# Add a web-hook for the app
+resource "heroku_app_webhook" "foobar_release" {
+  app_id  = "${heroku_app.foobar.id}"
+  level   = "notify"
+  url     = "https://example.com/heroku_webhook"
+  include = ["api:release"]
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `app_id` - (Required) The Heroku app to add to.
+* `level` - (Required) The webhook level (either `notify` or `sync`)
+* `url` - (Required) Optional plan configuration.
+* `include` - (Required) List of events to deliver to the webhook.
+* `secret` - (Optional) Value used to sign webhook payloads. Once set, this value cannot be fetched from the Heroku API, but it can be updated.
+* `authorization` - (Optional) Values used in `Authorization` header. Once set, this value cannot be fetched from the Heroku API, but it can be updated.
+
+## Importing
+
+Existing webhooks can be imported using the combination of the application name or id, a colon, and the webhook name or id, e.g.
+
+```
+$ terraform import heroku_app_webhook.foobar_release foobar:b85d9224-310b-409b-891e-c903f5a40568
+```
--- a/heroku/import_heroku_app_webhook_test.go
+++ b/heroku/import_heroku_app_webhook_test.go
@@ -1,0 +1,30 @@
+package heroku
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccHerokuAppWebhook_importBasic(t *testing.T) {
+	appName := fmt.Sprintf("tftest-%s", acctest.RandString(10))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+		},
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckHerokuAppWebhookConfig(appName, "https://terraform.example.com:4321", "sync", "api:build"),
+			},
+			{
+				ResourceName:        "heroku_app_webhook.foobar_webhook",
+				ImportStateIdPrefix: appName + ":",
+				ImportState:         true,
+			},
+		},
+	})
+}
--- a/heroku/resource_heroku_app_webhook.go
+++ b/heroku/resource_heroku_app_webhook.go
@@ -1,0 +1,236 @@
+package heroku
+
+import (
+	"context"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	validation "github.com/hashicorp/terraform/helper/validation"
+	heroku "github.com/heroku/heroku-go/v5"
+)
+
+func resourceHerokuAppWebhook() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceHerokuAppWebhookCreate,
+		Read:   resourceHerokuAppWebhookRead,
+		Update: resourceHerokuAppWebhookUpdate,
+		Delete: resourceHerokuAppWebhookDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: resourceHerokuAppWebhookImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"app_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"level": &schema.Schema{
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringInSlice([]string{"notify", "sync"}, true),
+			},
+
+			"url": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"include": &schema.Schema{
+				Type:     schema.TypeList,
+				Required: true,
+				MinItems: 1,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+					ValidateFunc: validation.StringInSlice([]string{
+						"api:addon-attachment",
+						"api:addon",
+						"api:app",
+						"api:build",
+						"api:collaborator",
+						"api:domain",
+						"api:dyno",
+						"api:formation",
+						"api:release",
+						"api:sni-endpoint",
+						"api:ssl-endpoint"}, true),
+				},
+			},
+
+			"secret": &schema.Schema{
+				Type:      schema.TypeString,
+				Optional:  true,
+				Sensitive: true,
+			},
+
+			"authorization": &schema.Schema{
+				Type:      schema.TypeString,
+				Optional:  true,
+				Sensitive: true,
+			},
+		},
+	}
+}
+
+// Callback for schema Resource.Create
+func resourceHerokuAppWebhookCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Config).Api
+
+	appId := getAppId(d)
+
+	opts := heroku.AppWebhookCreateOpts{
+		Level:   d.Get("level").(string),
+		URL:     d.Get("url").(string),
+		Include: getInclude(d),
+	}
+
+	if v, ok := d.GetOk("secret"); ok {
+		secret := v.(string)
+		opts.Secret = &secret
+	}
+
+	if v, ok := d.GetOk("authorization"); ok {
+		authorization := v.(string)
+		opts.Authorization = &authorization
+	}
+
+	webhook, err := client.AppWebhookCreate(context.TODO(), appId, opts)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(webhook.ID)
+
+	return nil
+}
+
+// Callback for schema Resource.Read
+func resourceHerokuAppWebhookRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Config).Api
+
+	appId := getAppId(d)
+
+	webhook, err := client.AppWebhookInfo(context.TODO(), appId, d.Id())
+	if err != nil {
+		return err
+	}
+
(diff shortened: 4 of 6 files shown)


mars:
Wow @CGA1123 😍 Thank y'all for contributing this resource!

`resource_heroku_webhook` is definitely a desirable feature. At first glance, it looks pretty complete. I'll exercise it and review the PR soon.

heroku/resource_heroku_webhook.go
```diff
@@ -0,0 +1,214 @@
...
+		Read:   resourceHerokuWebhookRead,
+		Update: resourceHerokuWebhookUpdate,
+		Delete: resourceHerokuWebhookDelete,
+
+		Schema: map[string]*schema.Schema{
+			"app": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
```
davidji99: Just a styling nitpick, will you add line breaks between each attribute definition?

heroku/resource_heroku_webhook.go
```diff
@@ -0,0 +1,214 @@
...
+
+func resourceHerokuWebhook() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceHerokuWebhookCreate,
+		Read:   resourceHerokuWebhookRead,
+		Update: resourceHerokuWebhookUpdate,
+		Delete: resourceHerokuWebhookDelete,
+
+		Schema: map[string]*schema.Schema{
+			"app": &schema.Schema{
```
davidji99: I'd like for new resources to define `app_id` instead of `app`. I know Heroku allows either the app name or the UUID but using the name causes issues like this #124. @mars, what do you think?

heroku/resource_heroku_webhook.go
```diff
@@ -0,0 +1,214 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"include": &schema.Schema{
+				Type:     schema.TypeList,
+				Required: true,
+				MinItems: 1,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+			"secret": &schema.Schema{
```
davidji99: This attribute is a good candidate for the `Sensitive: true` attribute property.

heroku/resource_heroku_webhook.go
```diff
@@ -0,0 +1,214 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"authorization": &schema.Schema{
```
davidji99: This attribute is a good candidate for the `Sensitive: true` attribute property.

heroku/resource_heroku_webhook_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+		if len(Webhook.Include) != 1 || Webhook.Include[0] != "api:release" {
+			return fmt.Errorf("Bad Include: %v", Webhook.Include)
+		}
+
+		return nil
+	}
+}
+
+// TODO:
```
davidji99: Are you planning on adding these tests in this PR?

heroku/resource_heroku_webhook.go
```diff
@@ -0,0 +1,214 @@
...
+			},
+			"level": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"url": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"include": &schema.Schema{
```
davidji99: If there are only two acceptable values for `include`, you can add this:

```
ValidateFunc: validation.StringInSlice([]string{"x', "y"}, true),
```

heroku/resource_heroku_webhook.go
CGA1123: While the `secret` and the `authorization` cannot be unset, they can be updated. In which case I'm not sure `ForceNew` is appropriate.

The state file and the remote cannot have a diff as these attributes are not accessible via any Heroku API!

heroku/resource_heroku_webhook.go
CGA1123: I didn't know about this, awesome!

For `include` there are 11 valid event types, do you think it is worth enumerating them and forcing that restriction?

I think it would be nice to enforce that correctness, however it would mean requiring a code change if/when Heroku add or remove any additional event types.

heroku/resource_heroku_webhook_test.go
CGA1123: Yes, I think these would be nice to have, although perhaps there is not that much value in them 🤔.

CGA1123:
@mars @davidji99 Thank you for taking looking at this so quickly! 😃

go.sum
```diff
@@ -203,6 +203,7 @@ github.com/hashicorp/serf v0.0.0-20160124182025-e4ec8cc423bb h1:ZbgmOQt8DOg796fi
 github.com/hashicorp/serf v0.0.0-20160124182025-e4ec8cc423bb/go.mod h1:h/Ru6tmZazX7WO/GDmwdpS975F019L4t5ng5IgwbNrE=
 github.com/hashicorp/terraform v0.12.6 h1:mWItQdLZQ7f3kBYBu2Kgdg+E5iZb1KtCq73V10Hmu48=
 github.com/hashicorp/terraform v0.12.6/go.mod h1:udmq5rU8CO9pEIh/A/Xrs3zb3yYU/W9ce1pp8K1ysHA=
+github.com/hashicorp/terraform v0.12.8 h1:51Z5K8oj42KZi0IiTf99wMTujDpvsTd3eaF1k5z1URk=
```
mars: This line creates a conflict, which actually reflects that this branch is behind master.

Master already includes a Terraform SDK v0.12.8 upgrade which was merged only a week ago, probably after you branched from master to make this PR.

Please rebase this branch on master.

heroku/resource_heroku_webhook.go
mars: I imagine that the Heroku API would return an error if the value is not allowable, but making this provider restrict values "at compile time" before making calls to Heroku API "at runtime" is an even better developer experience, failing faster.

It's simple enough to add more event types if/when Heroku releases them. Please follow @davidji99  suggestion and add this validation.

heroku/resource_heroku_webhook.go
mars: Yes please use `Sensitive: true` for `secret` attribute.

Agree that `ForceNew` is not apropos here, because the value is updatable.

heroku/resource_heroku_webhook.go
```diff
@@ -0,0 +1,252 @@
...
+	return nil
+}
+
+// Callback for schema Resource.Delete
+func resourceHerokuWebhookDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Config).Api
+	appId := getAppId(d)
+
+	// check if appId is valid
+	_, err := doesHerokuAppExist(appId, client)
```
mars: This extra API call seems unnecessary here, because delete will fail fine on its own.

heroku/resource_heroku_webhook.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+// Callback for schema Resource.Read
+func resourceHerokuWebhookRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Config).Api
+
+	appId := getAppId(d)
+
+	// check if appId is valid
+	_, err := doesHerokuAppExist(appId, client)
```
mars: This extra API call seems unnecessary here, because update will fail fine on its own.

heroku/resource_heroku_webhook.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+// Callback for schema Resource.Create
+func resourceHerokuWebhookCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Config).Api
+
+	appId := getAppId(d)
+
+	// check if appId is valid
+	_, err := doesHerokuAppExist(appId, client)
```
mars: Nice work checking for app existence first, but is this extra API call necessary? Wouldn't create fail fine on its own?

heroku/resource_heroku_webhook.go
```diff
@@ -0,0 +1,252 @@
...
+	// Enable Partial state mode and what we successfully committed
+	d.Partial(true)
+
+	client := meta.(*Config).Api
+	opts := heroku.AppWebhookUpdateOpts{}
+
+	appId := getAppId(d)
+
+	// check if appId is valid
+	_, err := doesHerokuAppExist(appId, client)
```
mars: Same as other `doesHerokuAppExist` usage; this seems unnecessary.

heroku/resource_heroku_webhook.go
```diff
@@ -0,0 +1,252 @@
...
+	return err
+}
+
+func resourceHerokuWebhookImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	client := meta.(*Config).Api
+
+	app, id, err := parseCompositeID(d.Id())
+
+	// check if app is valid
+	_, err = doesHerokuAppExist(app, client)
```
mars: Same as other `doesHerokuAppExist` usage; this seems unnecessary.

heroku/resource_heroku_webhook_test.go
mars: Of those three TODO, only the last one "test update" seems particularly useful. The other two would be testing Terraform core behavior.

website/docs/r/app_webhook.html.markdown
```diff
@@ -0,0 +1,39 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `app` - (Required) The Heroku app to add to.
+* `level` - (Required) The webhook level (either `notify` or `sync`)
+* `url` - (Required) Optional plan configuration.
+* `include` - (Required) List of events to deliver to the webhook.
+* `secret` - (Optional) Value used to sign webhook payloads.
+* `authorization` - (Optional) Values used in `Authorization` header.
```
mars: Would be nice for `secret` & `authorization` to include the note:

> Once set, this value cannot be fetched from the Heroku API, but it can be updated.

website/docs/r/app_webhook.html.markdown
```diff
@@ -0,0 +1,39 @@
+---
+layout: "heroku"
+page_title: "Heroku: heroku_webhook"
+sidebar_current: "docs-heroku-resource-webhook"
```
mars: This value should match the one set in **website/heroku.erb**, which is currently:

```suggestion
sidebar_current: "docs-heroku-resource-app-webhook"
```

website/docs/r/app_webhook.html.markdown
```diff
@@ -0,0 +1,39 @@
+---
+layout: "heroku"
+page_title: "Heroku: heroku_webhook"
```
mars: This value should match the resource name:

```suggestion
page_title: "Heroku: heroku_app_webhook"
```

website/docs/r/app_webhook.html.markdown
```diff
@@ -0,0 +1,39 @@
...
+## Example Usage
+
+```hcl
+# Create a new Heroku app
+resource "heroku_app" "default" {
+  name = "test-app"
+}
+
+# Add a web-hook for the app
+resource "heroku_webhook" "webhook" {
```
mars:
```suggestion
resource "heroku_webhook" "foobar_release" {
```

website/docs/r/app_webhook.html.markdown
```diff
@@ -0,0 +1,39 @@
...
+
+# heroku\_webhook
+
+Provides a [Heroku App Webhook](https://devcenter.heroku.com/categories/app-webhooks).
+
+## Example Usage
+
+```hcl
+# Create a new Heroku app
+resource "heroku_app" "default" {
```
mars: These resource names can be confusing to newcomers. Let's minimize confusion here as much as possible by showing a cohesive name relationship:

```suggestion
resource "heroku_app" "foobar" {
```

website/docs/r/app_webhook.html.markdown
```diff
@@ -0,0 +1,39 @@
...
+  name = "test-app"
```
mars:
```suggestion
  name = "foobar"
```

website/docs/r/app_webhook.html.markdown
```diff
@@ -0,0 +1,39 @@
+---
+layout: "heroku"
+page_title: "Heroku: heroku_webhook"
+sidebar_current: "docs-heroku-resource-webhook"
+description: |-
+  Provides the ability to manage application webhooks
+---
+
+# heroku\_webhook
```
mars:
```suggestion
# heroku\_app\_webhook
```

heroku/provider.go
```diff
@@ -87,6 +87,7 @@ func Provider() terraform.ResourceProvider {
 			"heroku_space_vpn_connection":              resourceHerokuSpaceVPNConnection(),
 			"heroku_team_collaborator":                 resourceHerokuTeamCollaborator(),
 			"heroku_team_member":                       resourceHerokuTeamMember(),
+			"heroku_app_webhook":                       resourceHerokuAppWebhook(),
```
davidji99: Move this below `heroku_app_release` instead of below `heroku_app_release`. Thanks!

website/docs/r/app_webhook.html.markdown
```diff
@@ -0,0 +1,39 @@
...
+
+```hcl
+# Create a new Heroku app
+resource "heroku_app" "foobar" {
+  name = "foobar"
+}
+
+# Add a web-hook for the app
+resource "heroku_webhook" "foobar_release" {
+  app     = "${heroku_app.default.name}"
```
davidji99: Rename this since the attribute was changed to `app_id`.

website/docs/r/app_webhook.html.markdown
```diff
@@ -0,0 +1,39 @@
...
+  url     = "https://example.com/heroku_webhook"
+  include = ["api:release"]
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `app` - (Required) The Heroku app to add to.
```
davidji99: Same as my comment above.

davidji99:
Once all the changes are done and settled, I recommend a bit of an interactive rebase to reduce the number of commits.

mars:
I just ran the full acceptance test suite on this PR, and got an error:

```
=== RUN   TestAccHerokuAppWebhook_importBasic
--- FAIL: TestAccHerokuAppWebhook_importBasic (9.58s)
    testing.go:569: Step 1 error: Resource specified by ResourceName couldn't be found: heroku_app_webhook.foobar
FAIL
```

CGA1123:
@mars Sorry about that. Forgot to check the import test 🤦‍♂
I've updated the test to look for the correct resource name (`foobar` -> `foobar_webhook`).

mars:
Hashicorp CI passed ✅

mars:
Here is the Terraform configuration that I used to try this out:
  https://github.com/mars/terraform-heroku-webhooks

Now merging & will request a new release.

mars:
This is now released in *v2.2.0*:

https://www.terraform.io/docs/providers/heroku/r/app_webhook.html

CGA1123:
Thanks 😃

PR #245, opened 2020-01-20
Remove quoted interpolation-only expressions

alexmuller:
As of Terraform 0.12 interpolation-only strings must be unquoted; they throw a deprecation warning:

> Warning: Interpolation-only expressions are deprecated


davidji99:
This provider is still backwards compatible for terraform 0.11, which still requires quotes for string interpolation. So a suggested change for your PR could be have a generic statement to inform the user to remove quotes for any string interpolation references in the docs.

Your change also brings up a question I'll ask Hashicorp on whether or not to drop the quotes in the docs right now.

Issue #93, opened 2018-07-16
heroku_addon_attachment uses app_id instead of app


ryanbrainard:
All the `heroku` resources that depend on `heroku_app` take `app` as a parameter; however, `heroku_addon_attachment` (added in #19) takes `app_id` instead. While this is easy to workaround, it is inconsistent with all the other resources. 

The same is true for the `addon_id` parameter, but probably less of an issue since it's less prevalent.

### Terraform Version
v0.11.7

### Affected Resource(s)
Please list the resources as a list, for example:
- `heroku_addon_attachment`

### Terraform Configuration Files
See below for relevent snippets

### Expected Behavior
Something like this should work:

```
resource "heroku_addon_attachment" "addons_kafka_01" {
  addon = "${heroku_addon.kafka_01.id}"
  app = "${heroku_app.addons_app.name}"
  name = "KAFKA_01"
}
```

### Actual Behavior
Instead, you have to do this:

```
resource "heroku_addon_attachment" "addons_kafka_01" {
  addon_id = "${heroku_addon.test.id}"
  app_id = "${heroku_app.addons_app.name}"
  name = "TEST_01"
}
```

### Steps to Reproduce
n/a

### Important Factoids
n/a

### References
#19 cc: @JamesBelchamber

JamesBelchamber:
The inconsistency is caused by what's available on the API - it's not actually possible to refer to the name of the app or the addon when creating the attachment, only the UUID. Check [the docs](https://godoc.org/github.com/cyberdelia/heroku-go/v3#AddOnAttachment) for details.

JamesBelchamber:
Nice! If this works for both the addon (`heroku_addon.test.name`) and app (`heroku_app.test.name`) then you should go ahead and submit a patch to the code, tests and docs :)

ryanbrainard:
@JamesBelchamber Do you know what this will mean for backward compatibility? I'm assuming we'll need to keep the `app_id` and `app_name` around as deprecated/hidden params as to not break existing configurations; however, will there be a path forward for users to upgrade? Can they re-import them or just do a refresh or something? Or would they just be stuck with the old version until they destroy and re-create the resource? I'm pretty new to Terraform, so not sure what's the best way to not break people but also give a path forward.

JamesBelchamber:
It'll break things :) I dunno if that's really a problem though. Someone with more authority will review it - personally I'd just break it and let someone tell you to make it backwards compatible (or not).

[...]

PR #255, opened 2020-03-07
Fixes in the datasource for heroku app

crisp2u:
This PR contains a fix for the datasource `heroku_app`
- fix id attribute of the app datasource (was returning the timestamp)
- add uuid attribute of the heroku app
- update and cleanup docs

### DIFF
--- a/heroku/data_source_heroku_app.go
+++ b/heroku/data_source_heroku_app.go
@@ -1,8 +1,6 @@
 package heroku
 
 import (
-	"time"
-
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
 
@@ -13,6 +11,11 @@ func dataSourceHerokuApp() *schema.Resource {
 			"name": {
 				Type:     schema.TypeString,
 				Required: true,
+			},
+
+			"id": {
+				Type:     schema.TypeString,
+				Computed: true,
 			},
 
 			"space": {
@@ -93,6 +96,10 @@ func dataSourceHerokuApp() *schema.Resource {
 					},
 				},
 			},
+			"uuid": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 		},
 	}
 }
@@ -106,7 +113,7 @@ func dataSourceHerokuAppRead(d *schema.ResourceData, m interface{}) error {
 		return err
 	}
 
-	d.SetId(time.Now().UTC().String())
+	d.SetId(app.App.Name)
 
 	if app.Organization {
 		err := setOrganizationDetails(d, app)
@@ -122,6 +129,7 @@ func dataSourceHerokuAppRead(d *schema.ResourceData, m interface{}) error {
 
 	d.Set("buildpacks", app.Buildpacks)
 	d.Set("config_vars", app.Vars)
+	d.Set("uuid", app.App.ID)
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


heroku/data_source_heroku_app.go
```diff
@@ -106,7 +113,7 @@ func dataSourceHerokuAppRead(d *schema.ResourceData, m interface{}) error {
 		return err
 	}
 
-	d.SetId(time.Now().UTC().String())
+	d.SetId(app.App.Name)
```
davidji99: Please set this to the app's UUID. The app's name can change but the UUID remains the same.

The reason behind this request has a bit of a background which you can read via https://github.com/terraform-providers/terraform-provider-heroku/issues/124 and https://github.com/terraform-providers/terraform-provider-heroku/issues/93.

crisp2u: I've read the discussion and I agree with the needed refactoring but for now I was trying to be consistent with the resource `heroku_app` which does the same thing [here](https://github.com/terraform-providers/terraform-provider-heroku/blob/392071b8b5cb028e690df5798e41bd46008b8f21/heroku/resource_heroku_app.go#L261)

I would say we should change this in a separate PR that will address also the resource, but my experience with Go not to mention Terraform Providers is very limited. 

If you decide to change the data source id now maybe I should drop also the UUID attribute then ?

heroku/data_source_heroku_app.go
```diff
@@ -15,6 +13,11 @@ func dataSourceHerokuApp() *schema.Resource {
 				Required: true,
 			},
 
+			"id": {
```
davidji99: Defining the `id` is not needed unless this attribute was used to lookup the resource remotely, which is taken care of by the `name` attribute above. You can drop this declaration in the schema.

heroku/data_source_heroku_app.go
davidji99: Totally understand your desire to be consistent with `heroku_app`. However, this means more 'tech debt' later on when the provider is refactored to prioritize the heroku app's UUID as the ID over the Name. So it would be great to not add to this 'debt'.

> If you decide to change the data source id now maybe I should drop also the UUID attribute then ?
Yeah, the proposed new `UUID` would be redundant when you set the resource id to its UUID.

heroku/data_source_heroku_app.go
crisp2u: Done

heroku/data_source_heroku_app_test.go
```diff
@@ -41,6 +43,8 @@ func TestAccDatasourceHerokuApp_Basic(t *testing.T) {
 						"data.heroku_app.foobar", "acm", "false"),
 					resource.TestCheckResourceAttr(
 						"data.heroku_app.foobar", "heroku_hostname", herokuHostname),
+					resource.TestCheckResourceAttrSet(
```
davidji99: You'll need to remove this since this resource attribute is no longer set.

heroku/data_source_heroku_app_test.go
```diff
@@ -25,6 +25,8 @@ func TestAccDatasourceHerokuApp_Basic(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(
 						"data.heroku_app.foobar", "name", appName),
+					resource.TestCheckResourceAttr(
+						"data.heroku_app.foobar", "id", appName),
```
davidji99: Can change this to use `TestCheckResourceAttrSet` since the ID app id is not known initially.

davidji99:
Looks good. Two additional minor changes and we should be good to go.

PR #261, opened 2020-04-15
Fix random VPN create failures

mars:
This test has been flappy for years. Let's fix it!

Ends-up, it's an issue with the VPN connection create process. Issuing the `POST` to create the VPN connection, and then immediately `GET` to retrieve status, results in occasional, seemingly random **not found** errors. This seems to be the same eventually-consistent behavior we observe with other resources of the Heroku Platform API.

Therefore, this changeset makes the VPN connection create method retry-await on **not found** as well as not-`active` statuses, until it eventually becomes `active`.

### DIFF
--- a/heroku/resource_heroku_space_vpn_connection.go
+++ b/heroku/resource_heroku_space_vpn_connection.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
@@ -79,6 +80,9 @@ func resourceHerokuSpaceVPNConnection() *schema.Resource {
 				},
 			},
 		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(45 * time.Minute),
+		},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


mars:
The VPN connection test passed in isolation, so now running full suite on this branch.

heroku/resource_heroku_space_vpn_connection.go
```diff
@@ -133,21 +137,25 @@ func resourceHerokuSpaceVPNConnectionCreate(d *schema.ResourceData, meta interfa
...
-			conn, err := client.VPNConnectionInfo(context.TODO(), space, id)
-			if err != nil {
-				return nil, "", fmt.Errorf("Error getting VPN status: %v", err)
-			}
-
-			return conn.ID, conn.Status, nil
-		},
-		Timeout: 45 * time.Minute,
-	}
+	err = resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
```
davidji99, line 140: Could we rename `err` to `retryErr`? Although, I don't believe the `err` in this line would conflict with the L141 `err`, more descriptive error variables would make a bit easier to understand this retry flow.

davidji99:
Looks good. I forgot about all the built-in retry functions the SDK has for providers!

I had one comment on variable names.

heroku/resource_heroku_space_vpn_connection.go
mars: For sure 👍

PR #263, opened 2020-04-23
[doc] Remove `name` from heroku_config example

ahivert:
```
Error: Unsupported argument

  on main.tf line 10, in resource "heroku_config" "common":
  10:     name = "common-vars"

An argument named "name" is not expected here.
```


mars:
Merging despite Acceptance test failure, because:

- the Acceptance tests will always fail for a forked repo ([secrets are not passed to the runner](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow))
- this change is simple documentation-only.

PR #268, opened 2020-05-29
There should be a data source for `heroku_pipeline`

### LINKED ISSUE #267: There should be a data source for `heroku_pipeline`
There should be a data source for `heroku_pipeline`.  If I want to separate heroku apps that share a pipeline into separate directories, they each end up managing the pipeline as a resource https://www.terraform.io/docs/providers/heroku/r/pipeline.html instead of just being able to reference and existing pipeline as a data source.

### Terraform and Heroku Provider Version

```
% terraform -v
Terraform v0.12.25
+ provider.heroku v2.4.1
+ provider.sumologic v2.0.1
```

### Expected Behavior

I would like to be able to reference an existing pipeline as data

```
data "heroku_pipeline" "mypipeline" {
  name = "mypipeline"
}

resource "heroku_app" "myapp" {
  pipeline = data.heroku_pipeline.mypipeline.id
}
```

davidji99:
Resolves #267

### DIFF
--- a/heroku/data_source_heroku_pipeline.go
+++ b/heroku/data_source_heroku_pipeline.go
@@ -1,0 +1,46 @@
+package heroku
+
+import (
+	"context"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceHerokuPipeline() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceHerokuPipelineRead,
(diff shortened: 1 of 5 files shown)


davidji99:
There is a limitation in the Heroku Platform API where a pipeline that has no apps cannot be retrieved using the name as the identifier, only its UUID. This poses some usage limitations on a data source for pipeline.

PR #273, opened 2020-06-23
Implement pipeline promotion

wchrisjohnson:
## Overview

A pipeline is a group of Heroku apps that share the same codebase. Each app in a pipeline represents one of the following stages in a continuous delivery workflow:

- Development
- Review
- Staging
- Production

See: https://devcenter.heroku.com/articles/pipelines

## Deployment

Pipelines let you define how your deployed code flows from one environment to the next. For example, you can deploy code to your staging app (which builds it into a slug) and later promote that same slug to production. This promotion flow ensures that production contains the exact same code that you tested in staging, and it’s also much faster than rebuilding the slug.

### Promoting

The Heroku CLI allows you to promote a given (source) app on a specific release within a pipeline to a downstage set of (target) apps:

https://devcenter.heroku.com/articles/pipelines#promoting

```
heroku pipelines:promote -a $SOURCE -r $STAGE -t $TARGET
```

This change adds the pipeline promotion feature from the Heroku CLI to the Heroku terraform provider.

### DIFF
--- a/heroku/config.go
+++ b/heroku/config.go
@@ -4,12 +4,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
-	"net"
 	"net/http"
 	"net/url"
 	"os"
 	"runtime"
-	"time"
 
 	"github.com/bgentry/go-netrc/netrc"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/logging"
@@ -56,14 +54,7 @@ func NewConfig() *Config {
 }
 
 func (c *Config) initializeAPI() (err error) {
-	transport := &http.Transport{
-		DialContext: (&net.Dialer{
-			Timeout: 5 * time.Second,
-		}).DialContext,
-	}
-
 	c.Api = heroku.NewService(&http.Client{
-		Timeout: 30 * time.Second,
 		Transport: &heroku.Transport{
 			Username: c.Email,
 			Password: c.APIKey,
@@ -71,15 +62,14 @@ func (c *Config) initializeAPI() (err error) {
 				heroku.DefaultUserAgent, version.ProviderVersion),
 			AdditionalHeaders: c.Headers,
 			Debug:             c.DebugHTTP,
-			Transport:         transport,
-			// Transport:         heroku.RoundTripWithRetryBackoff{
-			// Configuration fields for ExponentialBackOff
-			// InitialIntervalSeconds: 30,
-			// RandomizationFactor:    0.25,
-			// Multiplier:             2,
-			// MaxIntervalSeconds:     900,
-			// MaxElapsedTimeSeconds:  0,
-			// },
+			Transport:         heroku.RoundTripWithRetryBackoff{
+				// Configuration fields for ExponentialBackOff
+				// InitialIntervalSeconds: 30,
+				// RandomizationFactor:    0.25,
+				// Multiplier:             2,
+				// MaxIntervalSeconds:     900,
+				// MaxElapsedTimeSeconds:  0,
+			},
 		},
 	})
 
--- a/heroku/config.go
+++ b/heroku/config.go
@@ -4,10 +4,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
 	"runtime"
+	"time"
 
 	"github.com/bgentry/go-netrc/netrc"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/logging"
@@ -54,7 +56,14 @@ func NewConfig() *Config {
 }
 
 func (c *Config) initializeAPI() (err error) {
+	transport := &http.Transport{
+		DialContext: (&net.Dialer{
+			Timeout: 5 * time.Second,
+		}).DialContext,
+	}
+
 	c.Api = heroku.NewService(&http.Client{
+		Timeout: 30 * time.Second,
 		Transport: &heroku.Transport{
 			Username: c.Email,
 			Password: c.APIKey,
@@ -62,14 +71,15 @@ func (c *Config) initializeAPI() (err error) {
 				heroku.DefaultUserAgent, version.ProviderVersion),
 			AdditionalHeaders: c.Headers,
 			Debug:             c.DebugHTTP,
-			Transport:         heroku.RoundTripWithRetryBackoff{
-				// Configuration fields for ExponentialBackOff
-				// InitialIntervalSeconds: 30,
-				// RandomizationFactor:    0.25,
-				// Multiplier:             2,
-				// MaxIntervalSeconds:     900,
-				// MaxElapsedTimeSeconds:  0,
-			},
+			Transport:         transport,
+			// Transport:         heroku.RoundTripWithRetryBackoff{
+			// Configuration fields for ExponentialBackOff
+			// InitialIntervalSeconds: 30,
+			// RandomizationFactor:    0.25,
+			// Multiplier:             2,
+			// MaxIntervalSeconds:     900,
+			// MaxElapsedTimeSeconds:  0,
+			// },
 		},
 	})
 
--- a/heroku/resource_heroku_pipeline_promotion.go
+++ b/heroku/resource_heroku_pipeline_promotion.go
@@ -87,11 +87,6 @@ func resourceHerokuPipelinePromotionCreate(d *schema.ResourceData, meta interfac
 		log.Fatal("Error in create opts...")
 	}
 
-	// log.Printf("[DEBUG] PipelinePromotion create configuration: %#v", opts)
-	// log.Printf("--- PIPELINE ID: %s", opts.Pipeline.ID)
-	// log.Printf("--- SOURCE APP NAME: %s", *opts.Source.App.ID)
-	// log.Printf("--- TARGET APP NAME: %s", *opts.Targets[0].App.ID)
-
 	p, err := client.PipelinePromotionCreate(context.TODO(), opts)
 	if err != nil {
 		return fmt.Errorf("Error creating pipeline promotion: %s", err)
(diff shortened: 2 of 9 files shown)


sheax0r:
@wchrisjohnson Can you include some more info about what this is, and why and how to use it in the pull request comment? I'm trying to figure it out based on the tests, but some samples of what it does and the motiviation for it in a comment would be handy.

wchrisjohnson:
> @wchrisjohnson Can you include some more info about what this is, and why and how to use it in the pull request comment? I'm trying to figure it out based on the tests, but some samples of what it does and the motiviation for it in a comment would be handy.

@sheax0r Done - let me know what you think.

sheax0r:
@wchrisjohnson Thanks for the context Chris -- I'm actually also looking for more info about what it looks like to actually use this feature. Like, what does the terraform code look like, and what are the effects of applying it? 

"Promoting a pipeline" to me feels like a procedural operation, not like a declarative one. So I am trying to understand how this fits into a Terraform's declarative model.

wchrisjohnson:
@sheax0r the operation is a pipeline promotion, but it’s better conveyed as promoting a specific version/release of an app from one stage to another one. For example, in the Cedar sharding project, we have terraform that creates apps, but those apps are “empty”; there’s no cidecdeployed to them, they are not running, etc. 

Let’s say we have a version of ion (st01) running in staging, and we stand up a new prod shard, va01. In addition to creating the app, we need to deploy the latest code running on st01 to va01. This new resource enables us to do just that. 

I was thinking between the test and the Heroku docs I pointed to this would be crystal clear. Would it help if I added the above example and elaborated on it in the description?

heroku/config.go
```diff
@@ -54,22 +56,30 @@ func NewConfig() *Config {
...
-			Transport:         heroku.RoundTripWithRetryBackoff{
-				// Configuration fields for ExponentialBackOff
-				// InitialIntervalSeconds: 30,
-				// RandomizationFactor:    0.25,
-				// Multiplier:             2,
-				// MaxIntervalSeconds:     900,
-				// MaxElapsedTimeSeconds:  0,
-			},
+			Transport:         transport,
+			// Transport:         heroku.RoundTripWithRetryBackoff{
```
davidji99, line 75: Any reason to keep this commented code in?

heroku/resource_heroku_pipeline_promotion.go
```diff
@@ -0,0 +1,295 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	heroku "github.com/heroku/heroku-go/v5"
+)
+
+func resourceHerokuPipelinePromotion() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceHerokuPipelinePromotionCreate,
+		Read:   resourceHerokuPipelinePromotionRead,
+		Delete: resourceHerokuPipelinePromotionDelete,
+
```
davidji99, line 19: Is it possible to import an existing pipeline promotion resource? If not, it would be good to define the import function that returns a "not possible" error.

heroku/resource_heroku_pipeline_promotion.go
```diff
@@ -0,0 +1,295 @@
...
+	// Set the list of apps
+	if err := d.Set("targets", targets); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+//
+// Deeply nested Go structs are hard to grok, much less build. Here's
```
davidji99, line 156: Any reason to keep this commented code in?

davidji99:
I have the same sort of question that @sheax0r has regarding how someone would use this new resource.

In your test example, you have this HCL configuration:
```
resource "heroku_pipeline_promotion" "foobar-promotion" {
	pipeline = "${heroku_pipeline.foobar-pipeline.id}"
	source = "${heroku_app.foobar-source-app.uuid}"
	targets = ["${heroku_app.foobar-target-app1.uuid}","${heroku_app.foobar-target-app2.uuid}"]
}
```

Then, you briefly wrote this in your comment above (which you deleted):
> Let’s say we have a version of ion (st01) running in staging, and we stand up a new prod shard, va01. In addition to creating the app, we need to deploy the latest code running on st01 to va01. This new resource enables us to do just that.

It seems once you use `heroku_pipeline_promotion` to deploy the latest code (say 'v1') from `heroku_app.foobar-source-app.uuid` to `heroku_app.foobar-target-app1.uuid` and `heroku_app.foobar-target-app2.uuid`, I don't foresee a way to deploy 'v2' from the source app to the target apps using this resource. `heroku_pipeline_promotion`'s three user defined attributes, `pipeline`/`source`/`targets `, would remain the same so terraform won't recognize a `plan` diff and try to initiate a change.

----
> I was thinking between the test and the Heroku docs I pointed to this would be crystal clear. Would it help if I added the above example and elaborated on it in the description?

Yeah, the resource docs would be the perfect place to add additional examples and HCL examples (could be copied from your test).

heroku/config.go
wchrisjohnson: Good catch, @davidji99 - I apologize for not removing this before now.

We need to address this, but this PR isn't the place for it. Will revert.

heroku/resource_heroku_pipeline_promotion.go
wchrisjohnson: I have not considered it to date; my initial reaction is that there's not much value in it. I will take that suggestion and add a "noop" method.

heroku/resource_heroku_pipeline_promotion.go
wchrisjohnson: Yes, I added this comment after spending WAYYY more time than necessary navigating the right way. to build up the create opts struct.

wchrisjohnson:
@davidji99 From my perspective, Terraform is used to stand up resources, not necessarily manage versions of apps and other resources over time. You generally rely on CI/CD related systems to do those sorts of operations. I intentionally left out the Update and Delete operations on this specific resource because I think it will be used primarily to do the initial code deploy of one or more apps once those apps are created via Terraform.

I could be convinced to implement a basic delete of a PipelinePromotion when a `terraform destroy` is initiated, but I don't think terraform is generally used to manage, on an ongoing basis, specific versions/releases of apps in a production environment. Most of my understanding of the "sweet spot" of how Terraform is used is specific to the setup of resources, not fine grained management of apps and other resources as new releases are created.

If it would help, I can add this clarifying background to the description above.

wchrisjohnson:
@davidji99 I also neglected to add docs for the resource. Let me add those.

heroku/resource_heroku_pipeline_promotion.go
davidji99: Oh! This big comment block is for`createPipelinePromotionCreateOpts`.

davidji99:
> Terraform is used to stand up resources, not necessarily manage versions of apps and other resources over time.

> but I don't think terraform is generally used to manage, on an ongoing basis, specific versions/releases of apps in a production environment. Most of my understanding of the "sweet spot" of how Terraform is used is specific to the setup of resources, not fine grained management of apps and other resources as new releases are created.

Terraform could be used in both ways. For example in this provider, we have `heroku_app_release`, which manages the version of the running code for a dnyo. 

I wouldn't say it's limited to just standing up resources only. However, I agree that there are cases where a proposed resource isn't just a good fit for terraform. Looking back at my original comment of "how someone would use this new resource", I think I now get how this new resource could be used, which would be to standup the pipeline promotion resource. Declare that for this pipeline, this is the promotion path between the source app and the target app(s).

----

> I intentionally left out the Update and Delete operations on this specific resource because I think it will be used primarily to do the initial code deploy of one or more apps once those apps are created via Terraform.

Oh, it would be good to add a `Delete` operation. If I remove this resource after it's been applied and stored in state, I'd like to  see the promotion removed remotely. Given that all the user settable attributes are `ForceNew` at the moment, an `Update` operation is not needed and terraform will complain if you had added one.

sheax0r:
Hey Chris, 

I believe that what this resource accomplishes can also be accomplished with a `null_resource` and a `local-exec` provisioner. I'll elaborate with an example.

This:

```hcl
resource "heroku_pipeline_promotion" "foo" {
	pipeline = "${heroku_pipeline.foo.id}"
	source = "${heroku_app.source.uuid}"
	targets = ["${heroku_app.target.uuid}"]
}
```

Seems functionally equivalent to this:
```hcl
   triggers = {
     run_this_once = "run-this-once"
   }
  provisioner "local-exec" {
    command = "heroku pipelines:promote -a ${heroku_app.source.name} --to ${heroku_app.target.name}"
  }
```

This makes me think that this is not a good fit for a resource.

mars:
Thanks for opening this @wchrisjohnson 😄

How do you foresee triggering the promotion after its initial `terraform apply`?

What does the lifecycle look like?

wchrisjohnson:
> Thanks for opening this @wchrisjohnson 😄
> 
> How do you foresee triggering the promotion after its initial `terraform apply`?
> 
> What does the lifecycle look like?

@mars The lifecycle is pretty simple. As noted earlier, it would be used for the initial `pipeline:promote`, but not after that. I need to think a bit more about what @sheax0r is suggesting above, but his null_resource idea might be a better fit. If so, the only downside is all the time I've wasted implementing a resource that will end up getting deleted. 😞

wchrisjohnson:
> Seems functionally equivalent to this:
> 
> ```terraform
> resource "null_resource" "foo"
>   triggers = {
>     run_this_once = "run-this-once"
>   }
>   provisioner "local-exec" {
>     command = "heroku pipelines:promote -a ${heroku_app.source.name} --to ${heroku_app.target.name}"
>   }
> }
> ```

@sheax0r I've not called the Heroku CLI directly from terraform before. Is that just a dependency I need to make sure is available from my workstation? That feels brittle/problematic...

sheax0r:
@wchrisjohnson yes, you would need to ensure that the heroku CLI is available to the environment in which terraform runs.

mars:
@wchrisjohnson Based on how promotion works, this seems like an odd fit for a Terraform resource for two reasons:

1. the actual release promoted is completely time dependent (what commit is currently in the promoted stage) not based on declarative configuration, nor does it capture in state what was actually promoted
2. the inability to promote again without manually destroying or tainting the resource

I know it sucks to write code and then throw it out, but I am concerned that adding Pipeline promotion is an anti pattern for Terraform configuration.

davidji99:
> If so, the only downside is all the time I've wasted implementing a resource that will end up getting deleted. 😞

You're not alone on this! I've written two resources for this provider that I've nixed in the past because they just didn't fit the Terraform mold.

wchrisjohnson:
Closing this PR as the null_resource approach appears to be working. Thanks for the input.

PR #276, opened 2020-07-17
Update build & release process

davidji99:
In order to compile binaries that will be compatible with registry.terraform.io, we'll need to update our build scripts to generate release artifacts using `goreleaser`.

----
TODOs:
- [x] Update paths from `terraform-providers/terraform-provider-heroku` → `heroku/terraform-provider-heroku`
- [x] Bump `go.mod` Go version to `v1.14`.
- [x] Create release script
- [x] Introduce version.go


mars:
Nice @davidji99 👏

Looks like a search & replace for `terraform-providers/terraform-provider-heroku` → `heroku/terraform-provider-heroku` is needed to fix an acceptance test error.

Also, would you remove the TravisCI integration too?

davidji99:
> Looks like a search & replace for `terraform-providers/terraform-provider-heroku` → `heroku/terraform-provider-heroku` is needed to fix an acceptance test error.
> 
> Also, would you remove the TravisCI integration too?

Done!

davidji99:
@mars,

The high level release process explanation is first, we will track the semantic version in `version/version.go`. The new `build-release` script uses `version/version.go` version when creating a new git tag. To trigger `build-release`, one would run `make release` locally on their machine. There is validation in `build-release` to make sure that 1) the current local branch is `master`, 2) local/remote `master` are in sync, and 3) the new git tag doesn't already exist. If all the validations pass, then the `build-release` script creates and pushes up the new git tag. From there, `.github/workflow/release.yml` triggers and creates the artifacts and Github release. What do you think?

heroku/version.go
```diff
@@ -0,0 +1,16 @@
+// +build ignore
+
+// This a hack to populate the version in the custom binary file as this provider is not official.
```
mars, line 3: > this provider is not official

What does this mean? This is official from Heroku and we're a partner with Hashicorp.

version/version.go
```diff
@@ -1,13 +1,5 @@
...
-//https://github.com/terraform-providers/terraform-provider-azurerm/tree/master/version
-//This takes advantage of a new build flag populating the binary version of the
-//provider, for example:
-//-ldflags="-X=github.com/terraform-providers/terraform-provider-heroku/version.ProviderVersion=x.x.x"
-
 var (
-	// ProviderVersion is set during the release process to the release version of the binary, and
-	// set to acc during tests.
-	ProviderVersion = "dev"
+	ProviderVersion = "2.6.0"
```
mars, line 4: The previous way the version was passed in via linker flag kept it from being hardcoded in source code, and it was simply passed in from the tag-driven release process.

Why switch to hardcoding the version into source code?

If the release process is moved to use GitHub Releases, than the release tag can dynamically drive the binaries embedded version number.

scripts/build-release
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
mars, line 1: Why not use GitHub's Releases system, instead of building local tooling for it?

GitHub release process does everything this script does, plus offers Title, markdown Description, and integrates in many more ways through Actions.

mars:
Re-running CI to see if we get some fresh success (or failures)

heroku/version.go
davidji99: Oops, forgot to remove this.

version/version.go
davidji99: 👍  Will remove.

scripts/build-release
davidji99: Answered via DM.

PR #277, opened 2020-07-22
Update repo url in documentation

crisp2u:


mars:
We have a [draft pull request](https://github.com/heroku/terraform-provider-heroku/pull/276) in the works that includes all of the changes for the move Hashicorp Terraform Provider Registry.

The Provider documentation will be updated when our first release is completed using that new workflow.

PR #284, opened 2020-09-17
Add netrc support

sheax0r:
## Context

According to [GNU netrc documentation](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html#:~:text=netrc%20file-,The%20.,the%20command%20line%20option%20%2DN%20.), the `NETRC` environment variable can be used to specify a directory containing a `.netrc` file:

> The .netrc file contains login and initialization information used by the auto-login process. It generally resides in the user’s home directory, but a location outside of the home directory can be set using the environment variable NETRC.

Currently, this terraform provider does not honour the NETRC environment variable.

## Solution

Allow value of the `NETRC` environment variable to take precedence over the default, "~".

### DIFF
--- a/heroku/config.go
+++ b/heroku/config.go
@@ -164,13 +164,18 @@ func (c *Config) applyNetrcFile() error {
 	path := os.Getenv("NETRC_PATH")
 
 	if path == "" {
+		dir := os.Getenv("NETRC")
+		if dir == "" {
+			dir = "~"
+		}
+
 		filename := ".netrc"
 		if runtime.GOOS == "windows" {
 			filename = "_netrc"
 		}
 
 		var err error
-		path, err = homedir.Expand("~/" + filename)
+		path, err = homedir.Expand(fmt.Sprintf("%s/", dir) + filename)
 		if err != nil {
 			return err
 		}
--- a/heroku/config.go
+++ b/heroku/config.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
 	"runtime"
 
 	"github.com/bgentry/go-netrc/netrc"
@@ -175,7 +176,7 @@ func (c *Config) applyNetrcFile() error {
 		}
 
 		var err error
-		path, err = homedir.Expand(fmt.Sprintf("%s/", dir) + filename)
+		path, err = homedir.Expand(filepath.Join(dir, filename))
 		if err != nil {
 			return err
 		}
--- a/docs/index.md
+++ b/docs/index.md
@@ -100,6 +100,8 @@ machine api.heroku.com
 ...
 ```
 
+The directory containing the `.netrc` file can be overridden by the `NETRC` environment variable as described [here](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html).
+
 ## Argument Reference
 
 The following arguments are supported:


davidji99 (approved):

heroku/config.go
```diff
@@ -123,13 +123,18 @@ func (c *Config) applyNetrcFile() error {
...
+		}
+
 		filename := ".netrc"
 		if runtime.GOOS == "windows" {
 			filename = "_netrc"
 		}
 
 		var err error
-		path, err = homedir.Expand("~/" + filename)
+		path, err = homedir.Expand(fmt.Sprintf("%s/", dir) + filename)
```
bernerdschaefer, line 137: I'd suggest using [`filepath.Join`](https://golang.org/pkg/path/filepath/#Join) here, which will have better OS support than joining the file path strings manually.

website/docs/index.html.markdown
```diff
@@ -100,6 +100,8 @@ machine api.heroku.com
 ...
 ```
 
+The directory containing the `.netrc` file can be overridden by the `NETRC` environment variable as described [here](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html).
```
bernerdschaefer, line 103: > The .netrc file contains login and initialization information used by the auto-login process. It generally resides in the user’s home directory, but a location outside of the home directory can be set using the environment variable NETRC. Both locations are overridden by the command line option -N. The selected file must be a regular file, or access will be denied.

Have you tested the NETRC variable behavior with some tools to verify that it's meant to specify the _directory_ where the file is located? It's subtle, but my read of these docs is that it would specify the netrc _file_ location, and therefore function the same as the existing `NETRC_PATH` variable.

sheax0r: @bernerdschaefer I've been assuming that [the behaviour implemented by `heroku/netrc`](https://github.com/heroku/netrc/blob/master/lib/netrc.rb#L12) is correct, but I'll doublecheck this by seeing what `curl` does.

heroku/config.go
sheax0r: Works for me.

mars (approved):

website/docs/index.html.markdown
sheax0r: @bernerdschaefer I don't think this behaviour is implemented by curl. I think it's implemented by `ftp`. I don't have a good ftp server to test this out with :p. 

It is consistent with the `heroku/netrc` gem though as linked above, as well as `ftp` and some other stuff, eg [this](https://github.com/aio-libs/aiohttp/pull/2584/files). 

Do you have strong objections to this and/or alternative proposals? I'd like to merge this, but I'm open to other suggestions.

sheax0r: @bernerdschaefer If you'd prefer just to not have this in favour of the existing `NETRC_PATH` behaviour, I'm okay with it. In that case, I'd submit another pull request or mod this one to just describe that env var (in fact, I should add that anyway -- it's undocumented afaict).

bernerdschaefer: @sheax0r Awesome, thanks for looking into it. I'm good with the NETRC behavior added here! :shipit:

mikehale:
@sheax0r anything blocking this?

sheax0r:
@mikehale other than finding the time to circle back, rebase, verify and merge, I don't think so. I just don't have the time to finish it off. Feel free to take it over!

PR #288, opened 2020-11-04
Fix: heroku_app organization lock fails to modify setting

### LINKED ISSUE #287: heroku_app organization lock fails to modify setting
### Terraform Version

Terraform v0.12.26

### Heroku Provider Version

```
Terraform v0.12.26
+ provider.heroku v2.6.1
```

### Affected Resource(s)

- heroku_app

```hcl
resource "heroku_app" "app" {
  name   = "billmoritz-test-app"
  region = "virginia"
  space  = var.heroku_space

  organization {
    name = var.heroku_org
    locked = var.locked
  }
}

variable "heroku_org" {
    type = string
}

variable "heroku_space" {
    type = string
}

variable "locked" {
    type = string
}
```

### Debug Output

https://gist.github.com/billmoritz/2a053901b6dee2cb4a0577e9a2b2d666

### Panic Output

Panics don't happen unless I try to get debug level output by setting `TF_LOG=debug`

https://gist.github.com/billmoritz/2a053901b6dee2cb4a0577e9a2b2d666

### Expected Behavior

Changing the value of locked from false to true should update the app setting

### Actual Behavior

Nothing changes in Heroku and each time you plan it wants to change the setting. 

### Steps to Reproduce

1. `terraform apply -var 'locked=false' -var 'heroku_org=xxxxxxxxxxxxxxxxx' -var 'heroku_space=xxxxxxxxxxxxxxx'`
2. Verify value is set in console
3. `terraform apply -var 'locked=true' -var 'heroku_org=xxxxxxxxxxxxxxxxx' -var 'heroku_space=xxxxxxxxxxxxxxx'`
4. Verify value hasn't changed in console

davidji99:
Resolves #287

### DIFF
--- a/heroku/resource_heroku_app.go
+++ b/heroku/resource_heroku_app.go
@@ -145,12 +145,14 @@ func resourceHerokuApp() *schema.Resource {
 
 			"organization": {
 				Type:     schema.TypeList,
+				MinItems: 0,
+				MaxItems: 1,
 				Optional: true,
-				ForceNew: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"name": {
 							Type:         schema.TypeString,
+							ForceNew:     true,
 							Required:     true,
 							ValidateFunc: validation.StringIsNotEmpty,
 						},
@@ -164,6 +166,7 @@ func resourceHerokuApp() *schema.Resource {
 						"personal": {
 							Type:     schema.TypeBool,
 							Optional: true,
+							ForceNew: true,
 							Computed: true,
 						},
 					},
@@ -428,10 +431,12 @@ func resourceHerokuAppRead(d *schema.ResourceData, meta interface{}) error {
 	return nil
 }
 
+// resourceHerokuAppUpdate utilizes several unique API endpoints to update the app.
 func resourceHerokuAppUpdate(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*Config).Api
 	opts := heroku.AppUpdateOpts{}
 
+	// Make changes (if any) to the app itself.
 	if d.HasChange("name") {
 		v := d.Get("name").(string)
 		opts.Name = &v
@@ -447,6 +452,7 @@ func resourceHerokuAppUpdate(d *schema.ResourceData, meta interface{}) error {
 	}
 	d.SetId(updatedApp.Name)
 
+	// Make changes (if any) to the app's buildpack.
 	if d.HasChange("buildpacks") {
 		err := updateBuildpacks(d.Id(), client, d.Get("buildpacks").([]interface{}))
 		if err != nil {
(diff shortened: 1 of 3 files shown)


heroku/config.go
```diff
@@ -62,14 +62,6 @@ func (c *Config) initializeAPI() (err error) {
 				heroku.DefaultUserAgent, version.ProviderVersion),
 			AdditionalHeaders: c.Headers,
 			Debug:             c.DebugHTTP,
-			Transport:         heroku.RoundTripWithRetryBackoff{
```
davidji99, line 65: This was causing `panic: runtime error: invalid memory address or nil pointer dereference` when running `apply` or the acceptance tests. I'm removing it as we don't set any options here.

davidji99:
@mars Review please. Thanks!

heroku/resource_heroku_app.go
```diff
@@ -145,12 +145,14 @@ func resourceHerokuApp() *schema.Resource {
...
 				Type:     schema.TypeList,
+				MinItems: 0,
+				MaxItems: 1,
 				Optional: true,
-				ForceNew: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"name": {
 							Type:         schema.TypeString,
+							ForceNew:     true,
```
mars, line 155: 👍

heroku/resource_heroku_app.go
```diff
@@ -164,6 +166,7 @@ func resourceHerokuApp() *schema.Resource {
 						"personal": {
 							Type:     schema.TypeBool,
 							Optional: true,
+							ForceNew: true,
```
mars, line 169: 👍 nice correction adding these `ForceNew`.

heroku/config.go
mars: Oh no that's a bummer, but completely commenting it out removes the rate-limit retry/backoff behavior completely, which was implemented a few years ago to solve a problem for a rather large Heroku customer.

The commented values are there to show the optional attributes with their defaults.

Why did this start failing now? What changed since the last successful test runs? It's a standard HTTP RoundTripper, so unsure of it myself.

mars (commented):
Nice fix @davidji99 ! Just wondering about that incidental HTTP rate-limiting change…

heroku/config.go
davidji99: So if you `terraform apply` with `TF_LOG=DEBUG`, the `panic` occurs in the retrybackoff code in `heroku-go`. If you do't run in DEBUG mode, it's fine. It's weird.

I'll add it back for now. This was an unrelated issue anyhow.

mars: Noting it. Thank you @davidji99 ☺️

mars (approved):

PR #292, opened 2020-12-02
Ability to configure addon provisioning timeout

davidji99:
Often times, provisioning Heroku kafka takes longer than 20 minutes. Let's make the timeout configurable!

### DIFF
--- a/heroku/resource_heroku_addon.go
+++ b/heroku/resource_heroku_addon.go
@@ -127,11 +127,11 @@ func resourceHerokuAddonCreate(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	// Wait for the Addon to be provisioned
-	log.Printf("[DEBUG] Waiting for Addon (%s) to be provisioned", d.Id())
+	log.Printf("[DEBUG] Waiting for Addon (%s) to be provisioned", a.ID)
 	stateConf := &resource.StateChangeConf{
 		Pending: []string{"provisioning"},
 		Target:  []string{"provisioned"},
-		Refresh: AddOnStateRefreshFunc(client, app, d.Id()),
+		Refresh: AddOnStateRefreshFunc(client, app, a.ID),
 		Timeout: time.Duration(config.AddonCreateTimeout) * time.Minute,
 	}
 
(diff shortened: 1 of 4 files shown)


heroku/resource_heroku_addon.go
```diff
@@ -125,23 +126,24 @@ func resourceHerokuAddonCreate(d *schema.ResourceData, meta interface{}) error {
...
-		Timeout: 20 * time.Minute,
+		Timeout: time.Duration(config.AddonCreateTimeout) * time.Minute,
 	}
 
 	if _, err := stateConf.WaitForState(); err != nil {
 		return fmt.Errorf("Error waiting for Addon (%s) to be provisioned: %s", d.Id(), err)
 	}
 	log.Printf("[INFO] Addon provisioned: %s", d.Id())
 
+	// This should be only set after the addon provisioning has been fully completed.
```
mars, line 143: Failures in CI tests:

```
=== RUN   TestAccDatasourceHerokuAddon_Basic
Error:     testing.go:684: Step 0 error: errors during apply:
        Error: Error waiting for Addon () to be provisioned: timeout while waiting for state to become 'provisioned'
```

Looks like this change caused the above usage of `d.Id()` to become invalid, the retrier has no ID.

mars (changes requested):
Looks good, except the test failures.

heroku/resource_heroku_addon.go
davidji99: Ah! Oops!

heroku/resource_heroku_addon.go
```diff
@@ -125,23 +126,24 @@ func resourceHerokuAddonCreate(d *schema.ResourceData, meta interface{}) error {
...
+	d.SetId(a.ID)
+	log.Printf("[INFO] Addon ID: %s", d.Id())
```
mars, line 145: This seems to be logging the wrong ID, should it be `a.Id()`.

davidji99: That's okay because in L144, we set the resource `d` to `a.ID`.

mars: 👌  I see, looks good @davidji99 .

mars (approved):

heroku/resource_heroku_addon.go
davidji99: THanks!

PR #297, opened 2021-02-19
Update Tests

staaldraadtest:
Update test, test by @staaldraad


mars:
This looks like a malicious PR. It would disclose the Heroku API key to an external endpoint.

PR #299, opened 2021-03-04
Provide ability to not set `all_config_vars` on heroku_app resource

davidji99:
The `all_config_vars` can inadvertently set all non-terraform managed config vars
into the state file. This might fine for some but others may like the state file
to only include config var explicitly defined in a configuration. A solution here
is a provider block level attribute that will prevent the `heroku_app` resource from
setting this particular variable. In such a scenario, the attribute's value will be
an empty map.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -114,9 +114,16 @@ The following arguments are supported:
   provided, it will be sourced from the `HEROKU_HEADERS` environment variable
   (if set).
 
+* `customizations` - (Optional) Various attributes altering the behavior of certain resources.
+  Only a single `customizations` block may be specified, and it supports the following arguments:
+
+  * `set_app_all_config_vars_in_state` - (Optional) Controls whether the `heroku_app.all_config_vars` attribute
+    is set in the state file. Set to `false` if you'd like not to have non-managed config vars set in state.
+    Defaults to `true`.
+
 * `delays` - (Optional) Delays help mitigate issues that can arise due to
   Heroku's eventually consistent data model. Only a single `delays` block may be
-  specified and it supports the following arguments:
+  specified, and it supports the following arguments:
 
   * `post_app_create_delay` - (Optional) The number of seconds to wait after an
     app is created. Default is to wait 5 seconds.
--- a/heroku/config.go
+++ b/heroku/config.go
@@ -23,22 +23,28 @@ const (
 	DefaultPostDomainCreateDelay = int64(5)
 
 	// Default custom timeouts
-	DefaultAddonCreateTimeout = int64(20)
+	DefaultAddonCreateTimeout         = int64(20)
+	DefaultSetAppAllConfigVarsInState = true
 )
 
 type Config struct {
-	Api                   *heroku.Service
-	APIKey                string
-	DebugHTTP             bool
-	Email                 string
-	Headers               http.Header
+	Api       *heroku.Service
+	APIKey    string
+	DebugHTTP bool
+	Email     string
+	Headers   http.Header
+	URL       string
+
+	// Delays
 	PostAppCreateDelay    int64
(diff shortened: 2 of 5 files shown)


docs/index.md
```diff
@@ -114,9 +114,16 @@ The following arguments are supported:
   provided, it will be sourced from the `HEROKU_HEADERS` environment variable
   (if set).
 
+* `customizations` - (Optional) Various attributes altering the behavior of certain resources.
+  Only a single `customizations` block may be specified, and it supports the following arguments:
+
+  * `set_app_all_config_vars_in_state` - (Optional) Controls whether the `heroku_app.all_config_vars` attribute
+    is set in the state file. Set to `false` if you'd like not to have non-managed config vars set in state.
```
bernerdschaefer, line 121: Perhaps something like this to avoid the double-negative?

```suggestion
    is set in the state file. Set to `false` to only track managed config vars in the state.
```

It might also be helpful to include something explaining _why_ you'd want to change this behavior?

bernerdschaefer (commented):
Nice!

heroku/provider.go
```diff
@@ -36,6 +36,21 @@ func Provider() *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("HEROKU_API_URL", heroku.DefaultURL),
 			},
 
+			"customizations": {
```
mars, line 39: Why nest this inside a new list attribute? Why not `set_app_all_config_vars_in_state` as a top-level provider attribute?

heroku/resource_heroku_app_test.go
```diff
@@ -35,6 +35,39 @@ func TestAccHerokuApp_Basic(t *testing.T) {
...
+					resource.TestCheckResourceAttr(
+						"heroku_app.foobar", "name", appName),
+					resource.TestCheckResourceAttrSet(
+						"heroku_app.foobar", "uuid"),
+					resource.TestCheckResourceAttr(
+						"heroku_app.foobar", "config_vars.FOO", "bar"),
+					resource.TestCheckResourceAttr(
+						"heroku_app.foobar", "internal_routing", "false"),
+					resource.TestCheckResourceAttr(
+						"heroku_app.foobar", "all_config_vars.%", "0"),
```
mars, line 70: I imagine this is currently failing, because `set_app_all_config_vars_in_state` has not be changed to `false`.

heroku/config.go
```diff
@@ -103,10 +110,23 @@ func (c *Config) applySchema(d *schema.ResourceData) (err error) {
 		c.URL = url.(string)
 	}
 
+	if v, ok := d.GetOk("customizations"); ok {
+		vL := v.([]interface{})
+		if len(vL) > 1 {
+			return fmt.Errorf("Provider configuration error: only one customizations block is permitted")
+		}
+		for _, v := range vL {
+			delaysConfig := v.(map[string]interface{})
```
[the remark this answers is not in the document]
davidji99: Ah oops yeah!

heroku/resource_heroku_app_test.go
davidji99: Yeah, I plan on fixing this test. Just wanted to get the PR up for review.

docs/index.md
davidji99: 👍🏼

heroku/provider.go
davidji99: It could definitely be a top level provider attribute! However, I nested the attribute under `customizations` to keep in line with grouping the provider attributes under categories like `timeouts` and `delays`. If you have a better name for `customizations`, let me know!

bernerdschaefer (approved):

mars (approved):

PR #300, opened 2021-03-23
add instructions for importing domains by name

bpo:
This corrects the import instructions for domains, which use domain names rather than application UUIDs.

The example now also uses the same app/domain pair that is defined in HCL earlier in the document.

### DIFF
--- a/docs/resources/domain.md
+++ b/docs/resources/domain.md
@@ -44,8 +44,10 @@ The following attributes are exported:
 
 ## Importing
 
-When importing a Heroku domain resource, the ID must be built using the app name colon the unique ID from the Heroku API. For an app named `production-api` with a domain ID of `b85d9224-310b-409b-891e-c903f5a40568`, you would import it as: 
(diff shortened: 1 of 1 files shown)


docs/resources/domain.md
```diff
@@ -44,8 +44,8 @@ The following attributes are exported:
 
 ## Importing
 
-When importing a Heroku domain resource, the ID must be built using the app name colon the unique ID from the Heroku API. For an app named `production-api` with a domain ID of `b85d9224-310b-409b-891e-c903f5a40568`, you would import it as: 
+When importing a Heroku domain resource, the ID is specified `APP_NAME:DOMAIN_NAME`. For an app named `test-app` with a domain name of `terraform.example.com`, you would import it with:
```
davidji99, line 47: Both the domain name and UUID are acceptable here. You can use either domain identification so I would suggest a change here to reflect both possibilities.

davidji99 (changes requested):

docs/resources/domain.md
bpo: Aha, you are correct – I misunderstood the previous phrasing. Will fix.

davidji99 (approved):
Thanks!

PR #307, opened 2021-04-29
Remove deprecated heroku_space trusted_ip_ranges attribute

### LINKED ISSUE #306: heroku_space and heroku_space_inbound_ruleset resources are not compatible with each other.
**Terraform Version**
v0.14.8

### Heroku Provider Version
v4.2.0

### Affected Resource(s)
heroku_space and heroku_space_inbound_ruleset

### Expected Behavior
We are expecting one of these two.
1. heroku_space resource will not have the trusted-IP-ranges attribute, So only heroku_space_inbound_ruleset add the IP ranges in Heroku.
2. heroku_space resource by default leave the trusted-IP-ranges attribute empty, So trusted-IP ranges and allow list both are consistent.

### Actual Behavior
heroku_space resource trusted_ip_ranges attribute has been deprecated. But when heroku_space resource creates a private space it by default adds 0.0.0.0/0 in trusted IP ranges. And because of this in the subsequent run trusted_ip_ranges  try to override the allowlist. And because of this the trusted-IP-ranges and allow list are different and this cause of unexpected behavior in term of network-level access.

mars:
Fixes #306 by pruning out long-deprecated code that causes inconsistent results for `heroku_space_inbound_ruleset` resource.

### DIFF
--- a/heroku/resource_heroku_space.go
+++ b/heroku/resource_heroku_space.go
@@ -11,10 +11,9 @@ import (
 	heroku "github.com/heroku/heroku-go/v5"
 )
 
-type spaceWithRanges struct {
+type spaceWithNAT struct {
 	heroku.Space
-	TrustedIPRanges []string
-	NAT             heroku.SpaceNAT
+	NAT heroku.SpaceNAT
 }
 
 func resourceHerokuSpace() *schema.Resource {
@@ -74,17 +73,6 @@ func resourceHerokuSpace() *schema.Resource {
 				Default:  false,
 				ForceNew: true,
 			},
-
-			"trusted_ip_ranges": {
-				Type:       schema.TypeSet,
-				Computed:   true,
-				Optional:   true,
-				MinItems:   0,
-				Deprecated: "This attribute is deprecated in favor of heroku_space_inbound_ruleset. Using both at the same time will likely cause unexpected behavior.",
-				Elem: &schema.Schema{
-					Type: schema.TypeString,
-				},
-			},
 		},
 	}
 }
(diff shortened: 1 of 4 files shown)


davidji99:
I'll take a look @mars!

heroku/resource_heroku_space.go
```diff
@@ -74,17 +73,6 @@ func resourceHerokuSpace() *schema.Resource {
...
-			"trusted_ip_ranges": {
-				Type:       schema.TypeSet,
-				Computed:   true,
-				Optional:   true,
-				MinItems:   0,
-				Deprecated: "This attribute is deprecated in favor of heroku_space_inbound_ruleset. Using both at the same time will likely cause unexpected behavior.",
-				Elem: &schema.Schema{
-					Type: schema.TypeString,
-				},
-			},
```
davidji99, line 87: Do you know if an existing `heroku_space` resource that defines `trusted_ip_ranges` will trigger a destroy/recreate once the user upgrades to a version with this attribute removed? If so, it would be great to leverage a state migration here. You can see an example in the AWS provider or in `[herokux` provider](https://github.com/davidji99/terraform-provider-herokux/blob/master/herokux/resource_herokux_pipeline_github_integration_migrate.go).

mars: No, I don't 😞 

Maybe instead of deleting this code, we make it so the Space's Trusted IP Ranges code is completely ignored unless explicitly set? Seems like that would avoid the "phantom" API requests that cause the bad behavior.

bernerdschaefer: It looks like the official recommendation is to directly remove the attribute after deprecation: https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-removal.

mars: So, no migration required. @davidji99 would you take another look and if 👍, approve when you have a chance.

davidji99: @mars,

Are you planning on a major version bump then? I was hoping to avoid a major version bump.

davidji99 (approved):

PR #312, opened 2021-05-27
Use sni endpoint

mikehale:
There have been several changes around Heroku SSL and the ssl:endpoint add-on recently:
* Starting May 14 2021 new ssl:endpoints are no longer able to be provisioned: https://devcenter.heroku.com/changelog-items/2145
* Starting April 14 2021 the SSL Endpoint API should no longer be used for private space apps: https://devcenter.heroku.com/changelog-items/2095
* Starting March 22 2021 all apps have the option to have multiple certificates via Heroku SSL: https://devcenter.heroku.com/changelog-items/2082
* Starting in November 1 2021 the Domain Create endpoint will require a new param: https://devcenter.heroku.com/changelog-items/1938

Changes in this PR:
* It leaves the existing `heroku_cert` resource in place, but adds a note that it is deprecated. Should we do more there?
* It adds a new `heroku_ssl` resource. Multiple `heroku_ssl` resources can be specified per app.
* It adds a new optional `sni_endpoint_id` to `heroku_domain`. If you are using `heroku_ssl`, and want the certificate to be associated with a domain you must specify the `sni_endpoint_id` on the `heroku_domain` resource.

### DIFF
--- a/docs/resources/cert.md
+++ b/docs/resources/cert.md
@@ -19,10 +19,21 @@ resource "heroku_app" "default" {
   region = "us"
 }
 
-# Add-on SSL to application
-resource "heroku_addon" "ssl" {
-  app  = heroku_app.default.name
-  plan = "ssl"
+# Build a slug that we can scale
+resource "heroku_build" "default" {
+  app        = heroku_app.default.name
+  source {
+    url     = "https://github.com/mikehale/ruby-puma-getting-started/archive/v1.0.tar.gz"
+    version = "1.0"
+  }
+}
+
+# Scale the app to a tier that supports Heroku SSL
+resource "heroku_formation" "foobar-web" {
+  app = heroku_app.default.name
+  type = "web"
+  size = "hobby"
+  quantity = 1
 }
 
 # Establish certificate for a given application
@@ -47,7 +58,6 @@ The following arguments are supported:
 The following attributes are exported:
 
 * `id` - The ID of the add-on
-* `cname` - The CNAME for the SSL endpoint
 * `name` - The name of the SSL certificate
 
 ## Importing
--- a/docs/resources/cert.md
+++ b/docs/resources/cert.md
@@ -3,12 +3,12 @@ layout: "heroku"
 page_title: "Heroku: heroku_cert"
 sidebar_current: "docs-heroku-resource-cert"
 description: |-
-  Provides a Heroku SSL certificate resource. It allows to set a given certificate for a domain on a Heroku app.
+  Provides a Heroku SSL certificate resource. It allows to set a given certificate for a Heroku app.
 ---
 
 # heroku\_cert
 
-Provides a Heroku SSL certificate resource. It allows to set a given certificate for a domain on a Heroku app.
+Provides a Heroku SSL certificate resource. It allows to set a given certificate for a Heroku app.
 
 ## Example Usage
 
@@ -19,38 +19,18 @@ resource "heroku_app" "default" {
   region = "us"
 }
 
-# Build a slug that we can scale
-resource "heroku_build" "default" {
-  app = heroku_app.default.name
-  source {
-    url = "https://github.com/heroku/terraform-provider-heroku/raw/master/heroku/test-fixtures/app.tgz"
-  }
+# Add-on SSL to application
+resource "heroku_addon" "ssl" {
+  app  = heroku_app.default.name
+  plan = "ssl"
 }
 
-# Scale the app to a tier that supports Heroku SSL
-resource "heroku_formation" "web" {
-  app = heroku_app.default.name
-  type = "web"
-  size = "hobby"
-  quantity = 1
-  # Wait until the build has completed before attempting to scale
-  depends_on = ["heroku_build.default"]
-}
-
-# Create the certificate
+# Establish certificate for a given application
 resource "heroku_cert" "ssl_certificate" {
-  app = heroku_app.default.name
+  app               = heroku_app.default.name
   certificate_chain = file("server.crt")
-  private_key = file("server.key")
-  # Wait until the process_tier changes to hobby before attempting to create a cert
-  depends_on = ["heroku_formation.web"]
-}
-
-# Associate it with a domain
-resource "heroku_domain" "foobar" {
-  app = heroku_app.default.name
-  hostname = "terraform-123.example.com"
-  sni_endpoint = heroku_cert.ssl_certificate.id
+  private_key       = file("server.key")
+  depends_on        = ["heroku_addon.ssl"]
 }
 ```
 
@@ -66,12 +46,13 @@ The following arguments are supported:
 
 The following attributes are exported:
 
-* `id` - The ID of the SSL certificate
+* `id` - The ID of the add-on
+* `cname` - The CNAME for the SSL endpoint
 * `name` - The name of the SSL certificate
 
 ## Importing
 
-When importing a Heroku cert resource, the ID must be built using the app name colon the unique ID from the Heroku API. For an app named `production-api` with a certificate ID of `b85d9224-310b-409b-891e-c903f5a40568`, you would import it as: 
+When importing a Heroku cert resource, the ID must be built using the app name colon the unique ID from the Heroku API. For an app named `production-api` with a certificate ID of `b85d9224-310b-409b-891e-c903f5a40568`, you would import it as:
 
 ```
 $ terraform import heroku_cert.production_api production-api:b85d9224-310b-409b-891e-c903f5a40568
--- a/docs/resources/cert.md
+++ b/docs/resources/cert.md
@@ -3,12 +3,12 @@ layout: "heroku"
 page_title: "Heroku: heroku_cert"
 sidebar_current: "docs-heroku-resource-cert"
 description: |-
-  Provides a Heroku SSL certificate resource. It allows to set a given certificate for a Heroku app.
+  Provides a Heroku SSL certificate resource. It allows to set a given certificate for a domain on a Heroku app.
 ---
 
 # heroku\_cert
 
-Provides a Heroku SSL certificate resource. It allows to set a given certificate for a Heroku app.
+Provides a Heroku SSL certificate resource. It allows to set a given certificate for a domain on a Heroku app.
 
 ## Example Usage
 
@@ -37,13 +37,20 @@ resource "heroku_formation" "web" {
   depends_on = ["heroku_build.default"]
 }
 
-# Establish certificate for a given application
+# Create the certificate
 resource "heroku_cert" "ssl_certificate" {
-  app               = heroku_app.default.name
+  app = heroku_app.default.name
   certificate_chain = file("server.crt")
-  private_key       = file("server.key")
+  private_key = file("server.key")
   # Wait until the process_tier changes to hobby before attempting to create a cert
-  depends_on        = ["heroku_formation.web"]
+  depends_on = ["heroku_formation.web"]
+}
+
+# Associate it with a domain
+resource "heroku_domain" "foobar" {
+  app = heroku_app.default.name
+  hostname = "terraform-123.example.com"
+  sni_endpoint = heroku_cert.ssl_certificate.id
 }
 ```
 
@@ -59,7 +66,7 @@ The following arguments are supported:
 
 The following attributes are exported:
 
-* `id` - The ID of the add-on
+* `id` - The ID of the SSL certificate
 * `name` - The name of the SSL certificate
 
 ## Importing
--- a/docs/resources/cert.md
+++ b/docs/resources/cert.md
@@ -3,12 +3,12 @@ layout: "heroku"
 page_title: "Heroku: heroku_cert"
 sidebar_current: "docs-heroku-resource-cert"
 description: |-
-  Provides a Heroku SSL certificate resource. It allows to set a given certificate for a Heroku app.
+  Provides a Heroku SSL certificate resource. It allows to set a given certificate for a Heroku app. This resource is deprecated in favor of `heroku_ssl`.
 ---
 
 # heroku\_cert
 
-Provides a Heroku SSL certificate resource. It allows to set a given certificate for a Heroku app.
+Provides a Heroku SSL certificate resource. It allows to set a given certificate for a Heroku app.This resource is deprecated in favor of `heroku_ssl`.
 
 ## Example Usage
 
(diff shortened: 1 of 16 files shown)


davidji99:
The official Go documentation states that [tabs](https://golang.org/doc/effective_go#formatting) are used for indentation. Can you please run `go fmt` whenever the PR is ready to go so the indentation in the files you changed is reverted back to the original format? Thanks!

mikehale:
Apps which want to associate a certificate with a domain must specify which cert to use:

```
resource "heroku_cert" "ssl_certificate" {
  app = "${heroku_app.foobar.name}"
  certificate_chain="${file("cert.pem")}"
  private_key="${file("key.pem")}"
  depends_on = ["heroku_formation.foobar-web"]
}

resource "heroku_domain" "foobar" {
  app = "${heroku_app.foobar.name}"
  hostname = "terraform-123.example.com"
  sni_endpoint = heroku_cert.ssl_certificate.id
}
```

heroku/resource_heroku_cert_test.go
```diff
@@ -5,31 +5,17 @@ import (
...
 	"testing"
-	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 	heroku "github.com/heroku/heroku-go/v5"
 )
 
-// We break apart testing for EU and US because at present, Heroku deals with
```
mikehale, line 19: Since we are no longer dealing with ssl:endpoints, I don't think we need to test in US and EU anymore

docs/resources/cert.md
```diff
@@ -19,10 +19,20 @@ resource "heroku_app" "default" {
   region = "us"
 }
 
-# Add-on SSL to application
-resource "heroku_addon" "ssl" {
```
mikehale, line 23: TODO: this needs more updates

heroku/resource_heroku_domain.go
```diff
@@ -106,8 +134,18 @@ func resourceHerokuDomainRead(d *schema.ResourceData, meta interface{}) error {
...
 	return nil
 }
+
+func populateResource(d *schema.ResourceData, do *heroku.Domain) {
+	d.SetId(do.ID)
+	d.Set("app", do.App.Name)
+	d.Set("hostname", do.Hostname)
+	d.Set("cname", do.CName)
+	if v := do.SniEndpoint; v != nil {
+		d.Set("sni_endpoint", v.ID)
```
mikehale, line 149: TODO: do we want to use ID, or Name?

docs/resources/cert.md
davidji99: Just to make sure, does this file require more updates?

heroku/resource_heroku_cert_test.go
```diff
@@ -43,125 +29,60 @@ func TestAccHerokuCert_EU(t *testing.T) {
...
-	certFile := wd + "/test-fixtures/terraform.cert"
-	certFile2 := wd + "/test-fixtures/terraform2.cert"
-	keyFile := wd + "/test-fixtures/terraform.key"
-	keyFile2 := wd + "/test-fixtures/terraform2.key"
-
-	certificateChainBytes, _ := ioutil.ReadFile(certFile)
-	certificateChain := string(certificateChainBytes)
-	certificateChain2Bytes, _ := ioutil.ReadFile(certFile2)
-	certificateChain2 := string(certificateChain2Bytes)
+	slugID := testAccConfig.GetSlugIDOrSkip(t)
```
davidji99, line 32: Rather than requiring an existing slug id to execute this test, you can leverage the `heroku_slug` tests to create a slug on demand for the cert test(s) here: https://github.com/heroku/terraform-provider-heroku/blob/master/heroku/resource_heroku_slug_test.go#L204-L210

heroku/resource_heroku_domain.go
davidji99: Does the underlying API support both the sni_endpoint UUID and name?

Overall,  the UUID is generally preferred since it is idempotent. I would suggest to rename the attribute to be `sni_endpoint_id` to be clear that the attribute wants the UUID, not the name.

heroku/resource_heroku_domain_test.go
```diff
@@ -26,11 +28,85 @@ func TestAccHerokuDomain_Basic(t *testing.T) {
...
+		},
+	})
+}
+
+func TestAccHerokuDomain_SSL(t *testing.T) {
+	var domain heroku.Domain
+	var endpoint heroku.SniEndpoint
+	randString := acctest.RandString(10)
+	appName := fmt.Sprintf("tftest-%s", randString)
+	slugID := testAccConfig.GetSlugIDOrSkip(t)
```
davidji99, line 83: Same comment as https://github.com/heroku/terraform-provider-heroku/pull/312/files#r646300227

davidji99:
> * Starting May 14 2021 new ssl:endpoints are no longer able to be provisioned: https://devcenter.heroku.com/changelog-items/2145

If customers have already used a `heroku_cert` resource to provision an SSL endpoint, can these SSL endpoints still be updated and retrieved? If os, what should be the expected behavior when a user upgrades their provider to a version where `heroku_cert` is now managing the SNI endpoint?

docs/resources/cert.md
mikehale: thanks for the reminder, updated

heroku/resource_heroku_cert_test.go
mikehale: Updated in e7b51b1. I ended up copying the config into my test setup. Do you have a recommendation for a better way to do this? For instance are there ways to compose resource definitions?

heroku/resource_heroku_domain.go
```diff
@@ -48,38 +54,60 @@ func resourceHerokuDomainImport(d *schema.ResourceData, meta interface{}) ([]*sc
...
 	client := meta.(*Config).Api
-
 	app := d.Get("app").(string)
-	hostname := d.Get("hostname").(string)
+	opts := heroku.DomainCreateOpts{
+		Hostname: d.Get("hostname").(string),
+	}
+
+	if v := d.Get("sni_endpoint_id").(string); v != "" {
+		opts.SniEndpoint = &v
```
pafuent, line 77: Is it OK in Golang to take a reference of a variable that will be out of scope after the if ends?
I'm asking this because I'm not sure what will do Golang compiler/runtime in this case

heroku/resource_heroku_domain.go
```diff
@@ -48,38 +54,60 @@ func resourceHerokuDomainImport(d *schema.ResourceData, meta interface{}) ([]*sc
...
+}
+
+func resourceHerokuDomainUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Config).Api
+	app := d.Get("app").(string)
+	opts := heroku.DomainUpdateOpts{}
+
+	if d.HasChange("sni_endpoint_id") {
+		v := d.Get("sni_endpoint_id").(string)
+		opts.SniEndpoint = &v
```
pafuent, line 101: Same here

heroku/resource_heroku_domain.go
mikehale: Yes, for variables that are only used inside the conditional this is pretty common practice. An example: https://github.com/heroku/x/blob/023c823c2229dd9af1acbd354c0ea7758768df40/go-kit/metrics/testmetrics/provider.go#L56

mikehale:
> > * Starting May 14 2021 new ssl:endpoints are no longer able to be provisioned: https://devcenter.heroku.com/changelog-items/2145
> 
> If customers have already used a `heroku_cert` resource to provision an SSL endpoint, can these SSL endpoints still be updated? The changelog mentioned above indicates that "already provisioned SSL Endpoints will continue to work" and therefore the related API `GET` endpoints should continue but it isn't clear on whether existing SSL endpoints can be updated or not.
> 

Good question, yes updates to existing endpoints will still be supported.

> If so, what should be the expected behavior when a user upgrades their provider to a version where `heroku_cert` is now managing the SNI endpoint? Can an existing `heroku_cert` resource try to look up a non-existent SNI resource using the old SSL endpoint UUID? My question here pertains to the backwards compatibility and what is the upgrade path for existing `heroku_cert` resources as you have opted for a "replace things under the hood" vs creating a new resource.

The API `/apps/:id/ssl-endpoints` mean semantically different things for a private space app vs a common runtime app. Initially I was just thinking about the private space case which does not use an add-on, which is what lead me to update the existing resource. However, as you pointed out at least for now some customers will continue to use the add-on for common runtime apps.

I think it is probably worth trying to minimize customers needing to update configuration as much as possible, and it's also hard to think of another name that makes sense for managing a certificate, so I'll see how far I can get supporting the add-on common runtime case with the existing resource. If it proves too difficult/clever/tricky then we can revisit.

heroku/resource_heroku_cert_test.go
davidji99: What you have right now is good enough. We could do some more intricate string interpolations but this could become a more overarching refactor in a separate PR.

davidji99:
> I think it is probably worth trying to minimize customers needing to update configuration as much as possible, and it's also hard to think of another name that makes sense for managing a certificate, so I'll see how far I can get supporting the add-on common runtime case with the existing resource. If it proves too difficult/clever/tricky then we can revisit.

Sounds good! The testing part will become a bit tricky as we'd need to ensure that an existing `heroku_cert` exhibits a `noop` behavior once you implement your additional changes. This might be a difficult to simulate with the Terraform plugin SDK acceptance test framework.

mikehale:
> The testing part will become a bit tricky as we'd need to ensure that an existing `heroku_cert` exhibits a `noop` behavior once you implement your additional changes. This might be a difficult to simulate with the Terraform plugin SDK acceptance test framework.

Since we won't support creating ssl:endpoints anymore I was thinking I could use an app with an endpoint that I created outside the test, and reference it with something like `GetSSLEndpointAppIDOrAbort`. However, the tests complain that the app has already been created. The other thought I had was checking to see if the app had an ssl:endpoint add-on attached to it, and if it did to use that as the condition for which type of certificate to manage. My testing user still has access to create that add-on, but customers do not.

mikehale:
@davidji99 do you know if there is a way to see the list of dependancies for a resource?

mikehale:
I think we need to support both old and new style resources at the same time, so that people can migrate. In other words they'll start with a `heroku_cert` that is backed by an ssl:endpoint, and then need to add a `heroku_cert` that is backed by an sni endpoint, and point their domain at the new cert. I'm beginning to think it might be easier to have different resources for these. Maybe `heroku_cert` for the sni backed certs, and `heroku_cert_legacy` for ssl backed certs?

davidji99:
> @davidji99 do you know if there is a way to see the list of dependancies for a resource?

Can you elaborate further on this question? If I'm understanding correctly, I do not believe you can see the dependencies of one resource has on another resource via the SDK.

> I'm beginning to think it might be easier to have different resources for these. Maybe `heroku_cert` for the sni backed certs, and `heroku_cert_legacy` for ssl backed certs?

This approach is what I was hinting at earlier with all my questions around backwards compatibility and ease of use. As long as the documentation is up-to-date, there shouldn't be too much confusion between similarly named resources. I also suggest you keep `heroku_cert` as-is and create a new `heroku_cert_modern` resource. This way, existing users using `heroku_cert` will continue to work without issue.

mikehale:
> Can you elaborate further on this question? If I'm understanding correctly, I do not believe you can see the dependencies of one resource has on another resource via the SDK.

Yeah, that was my question... if I would be able to detect when the `heroku_cert` resource had a `depends_on` for `heroku_addon.ssl`.

mikehale:
> I also suggest you keep `heroku_cert` as-is and create a new `heroku_cert_modern` resource. This way, existing users using `heroku_cert` will continue to work without issue.

Makes sense. What do you think about `heroku_cert` vs `heroku_ssl`? Some other names for the new resource:
* `heroku_tls`
* `heroku_sni`
* `heroku_sni_endpoint`
* `heroku_multi_cert`

I kinda like `heroku_ssl` since that is what the [product](https://devcenter.heroku.com/articles/ssl) is called even if that doesn't match the names in the API.

davidji99:
> > I also suggest you keep `heroku_cert` as-is and create a new `heroku_cert_modern` resource. This way, existing users using `heroku_cert` will continue to work without issue.
> 
> Makes sense. What do you think about `heroku_cert` vs `heroku_ssl`? Some other names for the new resource:
> 
> * `heroku_tls`
> * `heroku_sni`
> * `heroku_sni_endpoint`
> * `heroku_multi_cert`
> 
> I kinda like `heroku_ssl` since that is what the [product](https://devcenter.heroku.com/articles/ssl) is called even if that doesn't match the names in the API.

`heroku_ssl` works for me.

mikehale:
@davidji99 thoughts on `sni_endpoint_id` vs `ssl_id` as the param on `heroku_domain`?
Any other thoughts or feedback in general? What do you think this needs to get it over the finish line? I'd love to see it shipped and customers start using it, so we can get folks off the old APIs.

davidji99:
> @davidji99 thoughts on `sni_endpoint_id` vs `ssl_id` as the param on `heroku_domain`?

Is this new attribute on `heroku_domain` reserved only for a value derived from `heroku_ssl.foobar.id`? Or can `heroku_cert.foobar.id` also work here? I guess it's the former, which then casts my vote for `ssl_id`.

> Any other thoughts or feedback in general? What do you think this needs to get it over the finish line? I'd love to see it shipped and customers start using it, so we can get folks off the old APIs.

I'll take an in-depth look today and leave any comments. We can target for a new provider release in the next day or so if no other pressing issues.

mikehale:
> Is this new attribute on `heroku_domain` reserved only for a value derived from `heroku_ssl.foobar.id`? Or can `heroku_cert.foobar.id` also work here? I guess it's the former, which then casts my vote for `ssl_id`.

Your assumption is correct. It will only work with a `heroku_ssl.foobar.id`.

> I'll take an in-depth look today and leave any comments. We can target for a new provider release in the next day or so if no other pressing issues.

Sounds good. I'm also happy to hand this off to you at this point if that works better.

davidji99:
Looks good @mikehale.

I saw a few tweaks to be made but I can do that in a followup PR tomorrow.

Otherwise, thanks for the work!

davidji99 (approved):

PR #313, opened 2021-06-18
Improvements to `heroku_ssl` resource

davidji99:
FYI @mikehale I made a few modifications to this resource:

- Rename `app` to `app_id`. There have been a few issues over the years of people complaining about app name changes causing unintentional downstream resource recreations. To address this going forward for new resources, I've been making sure the resource schema uses and validates the app UUID.
- I made `private_key` optional but clarified in the doc that it must be set during resource creation/modifications. The API unfortunately does not return the `private_key` in the JSON response. This makes things difficult during resource import as the next `apply` after `import` will register a `plan` diff indicating Terraform wishes to change `private_key` from `null --> <the_actual_value>`. I do not want to smush the private_key in the resource import ID either.

### DIFF
--- a/docs/resources/ssl.md
+++ b/docs/resources/ssl.md
@@ -82,8 +82,9 @@ The following arguments are supported:
 
 * `app_id` - (Required) The Heroku app UUID to add to.
 * `certificate_chain` - (Required) The certificate chain to add.
-* `private_key` - (Required) The private key for a given certificate chain.
-  This attribute value does not get displayed in logs or regular output.
+* `private_key` - (Optional) The private key for a given certificate chain. You **must** set this attribute when creating or
+  updating an SSL resource. However, **do not** set a value for this attribute if you are initially importing an existing
+  SSL resource. The attribute value does not get displayed in logs or regular output.
 
 ## Attributes Reference
 
--- a/docs/resources/ssl.md
+++ b/docs/resources/ssl.md
@@ -3,12 +3,16 @@ layout: "heroku"
 page_title: "Heroku: heroku_ssl"
 sidebar_current: "docs-heroku-resource-ssl"
 description: |-
-  Provides a Heroku SSL certificate resource. It allows to set a given certificate for a domain on a Heroku app.
+  Provides a Heroku SSL certificate resource to manage a certificate for a Heroku app.
 ---
 
 # heroku\_ssl
 
-Provides a Heroku SSL certificate resource. It allows to set a given certificate for a domain on a Heroku app.
+This resource manages an SSL certificate for a Heroku app.
+
+-> **IMPORTANT!**
+This resource renders the "private_key" attribute in plain-text in your state file.
+Please ensure that your state file is properly secured and encrypted at rest.
 
 ## Example Usage
 
@@ -33,15 +37,17 @@ resource "heroku_formation" "web" {
   type = "web"
   size = "hobby"
   quantity = 1
+
   # Wait until the build has completed before attempting to scale
   depends_on = [heroku_build.default]
 }
 
 # Create the certificate
 resource "heroku_ssl" "one" {
-  app = heroku_app.default.name
+  app_id = heroku_app.default.uuid
   certificate_chain = file("server.crt")
   private_key = file("server.key")
+
   # Wait until the process_tier changes to hobby before attempting to create a cert
   depends_on = [heroku_formation.web]
 }
@@ -55,7 +61,7 @@ resource "heroku_domain" "one" {
 
 # Create another certificate
 resource "heroku_ssl" "two" {
-  app = heroku_app.default.name
+  app_id = heroku_app.default.uuid
   certificate_chain = file("server.crt")
   private_key = file("server.key")
   # Wait until the process_tier changes to hobby before attempting to create a cert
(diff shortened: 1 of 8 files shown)


heroku/resource_heroku_ssl.go
```diff
@@ -11,17 +11,19 @@ import (
...
+		UpdateContext: resourceHerokuSSLUpdate,
+		DeleteContext: resourceHerokuSSLDelete,
 
 		Importer: &schema.ResourceImporter{
 			State: resourceHerokuSSLImport,
 		},
 
 		Schema: map[string]*schema.Schema{
-			"app": {
+			"app_id": {
```
davidji99, line 26: Standardizing new resources to use the `app_id` to prevent app renames from triggering dependent resource recreation.

heroku/resource_heroku_ssl.go
```diff
@@ -49,101 +51,103 @@ func resourceHerokuSSL() *schema.Resource {
...
 
 	d.SetId(ep.ID)
-	setErr := d.Set("app", app)
-	if setErr != nil {
-		return nil, setErr
-	}
+	d.Set("app_id", ep.App.ID)
+	d.Set("certificate_chain", ep.CertificateChain)
+	d.Set("name", ep.Name)
+	// TODO: need to add d.Set("private_key")
```
davidji99, line 68: TODO: Need to figure out how to set private_key on import or else `terraform` will detect a diff on the next `apply` after an `import`.

heroku/resource_heroku_ssl_test.go
```diff
@@ -62,30 +66,30 @@ resource "heroku_app" "one" {
...
   size = "hobby"
   quantity = 1
   depends_on = [heroku_app_release.one]
 }
 
 resource "heroku_ssl" "one" {
-  app = "${heroku_app.one.id}"
-  certificate_chain="${file("%s")}"
-  private_key="${file("%s")}"
+  app_id = heroku_app.one.uuid
```
mikehale, line 90: I don't think we have a uuid attribute, but the id is a uuid:

```suggestion
  app_id = heroku_app.one.id
```

docs/resources/ssl.md
```diff
@@ -33,15 +37,17 @@ resource "heroku_formation" "web" {
...
   quantity = 1
+
   # Wait until the build has completed before attempting to scale
   depends_on = [heroku_build.default]
 }
 
 # Create the certificate
 resource "heroku_ssl" "one" {
-  app = heroku_app.default.name
+  app_id = heroku_app.default.uuid
```
mikehale, line 47:
```suggestion
  app_id = heroku_app.default.id
```

docs/resources/ssl.md
```diff
@@ -55,7 +61,7 @@ resource "heroku_domain" "one" {
 
 # Create another certificate
 resource "heroku_ssl" "two" {
-  app = heroku_app.default.name
+  app_id = heroku_app.default.uuid
```
mikehale, line 64:
```suggestion
  app_id = heroku_app.default.id
```

docs/resources/ssl.md
```diff
@@ -74,9 +80,11 @@ resource "heroku_domain" "two" {
 
 The following arguments are supported:
 
-* `app` - (Required) The Heroku app to add to.
-* `certificate_chain` - (Required) The certificate chain to add
-* `private_key` - (Required) The private key for a given certificate chain
+* `app_id` - (Required) The Heroku app UUID to add to.
```
mikehale, line 83: Apps have an ID that is of type UUID: https://devcenter.heroku.com/articles/platform-api-reference#app

```suggestion
* `app_id` - (Required) The Heroku appID to add to.
```

docs/resources/ssl.md
```diff
@@ -87,7 +95,9 @@ The following attributes are exported:
 
 ## Importing
 
-When importing a Heroku ssl resource, the ID must be built using the app name colon the unique ID from the Heroku API. For an app named `production-api` with a certificate ID of `b85d9224-310b-409b-891e-c903f5a40568`, you would import it as:
+An existing SSL resource can be imported using a composite value of the app name and certificate UUID separated by a colon.
```
mikehale, line 98:
```suggestion
An existing SSL resource can be imported using a composite value of the app name and certificate ID separated by a colon.
```

mikehale:
@davidji99 Thanks for these improvements. One thing I noticed is the use of `app.uuid` which is not something provided by the API. We have an `app.id`, which is of type `uuid`. I noticed a few other places in the code and docs on different resources like App, Build, or Pipeline that should probably be updated from UUID to ID.

heroku/resource_heroku_ssl_test.go
davidji99: `uuid` is computed attribute of `heroku_app`: https://github.com/heroku/terraform-provider-heroku/blob/master/heroku/resource_heroku_app.go#L176-L178

And `heroku_app.id` is set to the name of the app: https://github.com/heroku/terraform-provider-heroku/blob/master/heroku/resource_heroku_app.go#L321

Oh the joys of two primary keys.

docs/resources/ssl.md
davidji99: See https://github.com/heroku/terraform-provider-heroku/pull/313#discussion_r655698240

docs/resources/ssl.md
davidji99: See https://github.com/heroku/terraform-provider-heroku/pull/313#discussion_r655698240

heroku/resource_heroku_ssl_test.go
mikehale: oh, wow

mikehale (approved):

heroku/resource_heroku_ssl_test.go
davidji99: Yeah, there has been an inconsistency in the provider over what primary key is used as the resource ID. Fixing this consistency would require a some work around adding migrations so users don't end up with mass resource recreations.

For new resources, I've been steering us towards the path of setting `resource.id` to the UUID and then having `resource.name` for convenience.

PR #314, opened 2021-06-23
Fix heroku_domain so sni_endpoint_id is not cleared on update

brianhelba:
Currently, when `sni_endpoint_id` is not provided by the configuration, the call to `d.HasChange("sni_endpoint_id")` will always determine that a change has occurred, as the API will return the actual value from `do.SniEndpoint.ID` but the "desired" configuration value will be unset. Thus in this case, updates will always attempt to modify Heroku to set the value to the empty string, which Heroku disallows (and is semantically incorrect).

By making `sni_endpoint_id` both `Optional` and `Computed`, it can still be explicitly set via configuration, but otherwise Terraform use the value from creation (which is automatically determined by Heroku) as the set value for updates. Since this will match the current value from the API, no update will be attempted, which is the correct behavior.

### DIFF
--- a/heroku/resource_heroku_domain.go
+++ b/heroku/resource_heroku_domain.go
@@ -42,6 +42,7 @@ func resourceHerokuDomain() *schema.Resource {
 			"sni_endpoint_id": {
 				Type:     schema.TypeString,
 				Optional: true,
+				Computed: true,
 			},
 		},
 	}


brianhelba:
This fixes a regression caused by #312. Until this is merged and released, `heroku_domain` is broken on the latest release.

@davidji99 @mikehale Please take a look.

davidji99:
Hi @brianhelba,

Thanks for the PR. Do you have sample Terraform configuration code that would allow one to replicate the behavior you are seeing below:

> updates will always attempt to modify Heroku to set the value to the empty string

Currently, I am unable to replicate this issue with a basic Terraform configuration below:

```hcl
  name = "test-sample"
  region = "us"
}

resource "heroku_domain" "foobar" {
  app = heroku_app.foobar.id
  hostname = "terraform-test-sample.example.com"
}
```

If I `apply` the above, the state file has `"sni_endpoint_id": null`. Then, a subsequent `plan` will not detect a configuration change. Based on your observations, I would expect to see the below in a `plan` diff when `sni_endpoint_id` is not set for a `heroku_domain` resource:

```
- sni_endpoint_id = "some_UUID" -> null
```

I have no objections to adding `Computed` to `sni_endpoint_id`. However, I'd like to know the configuration code so we can add the code as an acceptance test for the resource.

mikehale:
I was able to replicate the issue. I think this is connected with the auto-association behavior which will be [sunset this fall](https://devcenter.heroku.com/changelog-items/2192).

Currently the first `heroku_ssl` resource on an app will automatically attempt to connect itself with any domains that already exist on the app. For most folks this won't be a problem because usually when you have a `heroku_ssl` (cert) you want to associate it with a domain, and specifying the association effectively bypasses the auto-association, keeping heroku and terraform in sync

```
resource "heroku_ssl" "one" {
  app_id            = heroku_app.one.uuid
  certificate_chain = file("heroku/test-fixtures/terraform.cert")
  private_key       = file("heroku/test-fixtures/terraform.key")
  depends_on        = [heroku_formation.web]
}

resource "heroku_domain" "default" {
  app        = heroku_app.one.name
  hostname   = "mhale-terraform-test-1.example.com"
  sni_endpoint_id = heroku_ssl.one.id
}
```

The issue arrises if you don't want to associate it with a domain, and the auto-association logic kicks in:

```
resource "heroku_ssl" "one" {
  app_id            = heroku_app.one.uuid
  certificate_chain = file("heroku/test-fixtures/terraform.cert")
  private_key       = file("heroku/test-fixtures/terraform.key")
  depends_on        = [heroku_formation.web]
}

resource "heroku_domain" "default" {
  app        = heroku_app.one.name
  hostname   = "mhale-terraform-test-1.example.com"
}
```
... results in a difference in the plan on a subsequent run:

```
sni_endpoint_id = "some_UUID" -> null
```

If you want to have a `heroku_ssl` resource that is not associated with any domains you need to make sure the domains are created afterwards by adding a depends:

```
resource "heroku_ssl" "one" {
  app_id            = heroku_app.one.uuid
  certificate_chain = file("heroku/test-fixtures/terraform.cert")
  private_key       = file("heroku/test-fixtures/terraform.key")
  depends_on        = [heroku_formation.web]
}

resource "heroku_domain" "default" {
  app        = heroku_app.one.name
  hostname   = "mhale-terraform-test-1.example.com"
  depends_on = [heroku_ssl.one]
}
```

brianhelba:
In my use case, I'm using exclusively a `heroku_domain`, with no `heroku_ssl`. My `heroku_app` has `acm = True`. My full configuration is here, if it helps: https://github.com/girder/terraform-heroku-django/blob/master/modules/heroku/main.tf

> ... results in a difference in the plan on a subsequent run:
> 
> ```
> sni_endpoint_id = "some_UUID" -> null
> ```
This is precisely the problematic behavior I've experienced. When these changes get applied on a subsequent update, it fails because it attempts to send a `PATCH` to the Heroku API with `sni_endpoint=""`.

mikehale:
Ah, `acm=true`. That would do it. Will have to think about this.

brianhelba:
@mikehale Thanks.

I will have to pin to `< v4.5.0`. `acm = True` was working fine before the latest release.

mikehale:
> I have no objections to adding `Computed` to `sni_endpoint_id`. However, I'd like to know the configuration code so we can add the code as an acceptance test for the resource.

I think it will be difficult to test the acm case, because in addition to setting `acm=true` we would need to fetch the domain target cname, and update DNS with that value so that the ACM process can validate, and create the LE certificate and associated SniEndpoint.

I'm also wondering what is the downside to having the attribute be computed. Would that cause us to miss configuration changes that we would like to know about? Should we only ignore it when `acm=true`?

davidji99:
Closing this PR in favor of #316.

PR #318, opened 2021-07-01
Ability for `heroku_team_collaborator` to handle users that already exist on the app

davidji99:
From my experience, people often manually add users to Heroku team apps outside of Terraform for a number of reasons. When the Terraform configuration is updated to reflect such out-of-band changes, people need to `import` the resource first before executing an `apply`. This PR will handle this scenario but checking for a specific error returned from resource creation indicating that the target user is on the target app. If so, find the collaborator ID, set it to the resource ID, and proceed with state refresh.


davidji99:
@mars Thoughts on this?

mars (approved):
This looks good @davidji99 👍  I really appreciate these little adaptations to make Terraform cooperate with external configuration changes.

PR #319, opened 2021-07-14
Source build checksum

### LINKED ISSUE #289: heroku_build local_checksum constantly changing
### Terraform Version

v0.12.x

### Heroku Provider Version

3.0.0

### Affected Resource(s)

- heroku_build

### Expected Behavior

Using `heroku_build`'s `path` attribute with a directory should only trigger a new build if the contents of the files in the directory change.

### Actual Behavior

We use terraform enterprise, which means that each time we run our code, the files will have different metadata (ex: modified time).  This means that the tarball generated by the provider will have a different checksum on each run.

### Proposal

I propose (and am willing to contribute) that when `heroku_build`'s `path` is set to a directory that the checksum be calculate before building the tarball and that we write the code such that metadata like modified time are not included in the checksum.

### References

- #269 (note some comments about using terraform cloud that fit with what we are seeing)

mars:
Fixes #289 

This new `heroku_build` "relaxed checksum" strategy only effects [builds with `source.path`](https://registry.terraform.io/providers/heroku/heroku/latest/docs/resources/build#local-source) pointing to a **directory**.

In this case, the checksum is a now non-standard algorithm, only hashing the file paths & contents, not the actual tarball uploaded to Heroku. This should allow `heroku_build` in ephemeral Terraform runtimes, such as Terraform Cloud, to only rebuild when source files actually change. It's possible this diffing has become too relaxed… real-world testing will inform whether this change really works well.

Also, the relaxed checksum is not set on the Heroku build, because it is non-standard and will fail the build. It is only saved in Terraform state as the `local_checksum` attribute.

-----

I've tagged a pre-release to test this. Hopefully it builds!
https://github.com/heroku/terraform-provider-heroku/releases/tag/v4.6.0-pre1


heroku/resource_heroku_build.go
```diff
@@ -271,6 +282,8 @@ func resourceHerokuBuildCreate(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	d.SetId(build.ID)
+	// Capture the checksum, to diff changes in the local source directory.
+	d.Set("local_checksum", checksum)
```
davidji99, line 286: We need to still set `local_checksum` to a value returned from the API like this removed [line](https://github.com/heroku/terraform-provider-heroku/pull/319/files#diff-198a71a15d2e3540b5111c003cf9b0ba9ca0872287516b25b581cb930776da6cL452) when `heroku_build.source` is set to a file path. Right now, this line sets `local_checksum` to whatever was locally calculated without confirming this checksum with the Platform API.

My suggestion is to pipe the value of `useRelaxedChecksum` through `resourceHerokuBuildRead` to `setBuildState` so that you can add back this [line](https://github.com/heroku/terraform-provider-heroku/pull/319/files#diff-198a71a15d2e3540b5111c003cf9b0ba9ca0872287516b25b581cb930776da6cL452). You achieve this by setting this [line](https://github.com/heroku/terraform-provider-heroku/pull/319/files#diff-198a71a15d2e3540b5111c003cf9b0ba9ca0872287516b25b581cb930776da6cR299) to call `resourceHerokuBuildRead` instead of `return nil`.

davidji99 (changes requested):
Looks good. I have a suggested change around setting `local_checksum` in state.

heroku/resource_heroku_build.go
mars: We cannot set the relaxed checksum on the Heroku Build. It is a non-standard checksum that Heroku doesn't understand, and Heroku fails the build with `Checksum mismatch` error. The standard tarball checksum is still applied for builds that use a pre-made tarball; the `heroku_build` acceptance tests validate this.

Heroku API doesn't produce a checksum if it's not set when the build is created. The checksum is only read by Heroku build to verify the source tarball before the build is executed. The checksum is meaningless in this situation anyway, since it's a one-off tarball that is not archived or reused. We can trust that TCP/IP will send all the bytes correctly.

By using exclusively `local_checksum` to maintain state for these source directory builds, this fixes the outstanding problem.

So, unless we discover some bug or use-case that breaks with this Terraform-local relaxed checksum, it should remain only in `local_checksum`.

mars:
@davidji99 it looks like there's a new problem with the `heroku_app_config_association` in acceptance tests:

```
=== RUN   TestAccHerokuAppConfigAssociation_Advanced
Error:     resource_heroku_app_config_association_test.go:40: Step 1/1 error: Error retrieving state after apply: exit status 2
        panic: value is marked, so must be unmarked first
        
        goroutine 1 [running]:
        github.com/zclconf/go-cty/cty.Value.assertUnmarked(...)
          /home/circleci/go/pkg/mod/github.com/zclconf/go-cty@v1.8.4/cty/marks.go:141
```

The panic stack doesn't include any of our provider code 🙈  and I cannot reproduce it running `make testacc` locally. Re-running the test suite to see if it's flappy.

mars:
Ah, I just found that our GitHub Action acceptance test suite is using a pre-installed version of `terraform`, [currently `v1.0.1`](https://github.com/actions/virtual-environments/blob/ubuntu20/20210628.1/images/linux/Ubuntu2004-README.md). This bug is fixed in `v1.0.2`.

I'm going to update the acceptance test action to download and use an explicitly set version of `terraform`, based on the [Running Acceptance Tests docs](https://www.terraform.io/docs/extend/testing/acceptance-tests/index.html#running-acceptance-tests).

davidji99 (approved):

PR #323, opened 2021-12-01
Modify heroku_drain to accept URLs that pass in basic auth

davidji99:
It is possible to define a log drain URL that has a username and password: https://devcenter.heroku.com/articles/log-drains#https-drains<br><br>We need to modify heroku_drain to be able to safely accept such a value.

### DIFF
--- a/heroku/resource_heroku_drain.go
+++ b/heroku/resource_heroku_drain.go
@@ -23,16 +23,26 @@ func resourceHerokuDrain() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
-			"url": {
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+			"app": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				AtLeastOneOf: []string{"url", "sensitive_url"},
 			},
 
-			"app": {
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+			"url": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"sensitive_url"},
+			},
+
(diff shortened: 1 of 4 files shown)


heroku/resource_heroku_drain.go
```diff
@@ -48,9 +57,27 @@ const retryableError = `App hasn't yet been assigned a log channel. Please try a
...
-		return nil, err
+	result := strings.Split(d.Id(), ":")
+
+	var app, id string
+	var isSensitive bool
+	switch len(result) {
+	case 2:
+		app = result[0]
+		id = result[1]
+		isSensitive = false
```
sheax0r, line 68: I believe you could still have a sensitive 2-string basic auth URL with just a user name:

```
https://some-long-random-username@mylogdrain.com
```

mars: Looks like this is parsing the ID for import, not the actual drain URL @sheax0r .

heroku/resource_heroku_drain.go
```diff
@@ -107,6 +154,10 @@ func resourceHerokuDrainDelete(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("Error deleting drain: %s", err)
 	}
 
+	log.Printf("[INFO] Deleted drain: %s", d.Id())
+
+	d.SetId("")
```
mars, line 159: This is not necessary when deleting. Returning `nil` (no error), is all that is required to delete successfully, removing from Terraform state.

mars (approved):
Looks great @davidji99 😄👍

heroku/resource_heroku_drain.go
davidji99: All good to know @mars.

heroku/resource_heroku_drain.go
sheax0r: My bad :)

PR #324, opened 2021-12-10
Fix `heroku_addon` update

mars:
In response to a Heroku-internal issue with `heroku_addon` update,

* add test for updating an add-on's plan & config
* fix to update add-on's config during update.


mars:
Closing this, because add-on config cannot be updated by the user (only the add-on provider):

[Add-on Config Update](https://devcenter.heroku.com/articles/platform-api-reference#add-on-config-update)
> Update an add-on’s config. Can only be accessed by the add-on partner providing this add-on.

PR #325, opened 2021-12-15
heroku_addon "config_var_values"

mars:
Currently, `heroku_addon` config vars are not readable from their parent `heroku_app`, until the next Terraform refresh.

You've probably encountered this problem, if you've ever tried to set a Terraform `output` to the value of an add-on config var, or tried to use an add-on's config vars to operate on an add-on immediately after it's created.

This PR remedies this problem by presenting each add-on's config var values directly on the add-on resource itself as a new, sensitive `config_var_values` map attribute.

Using this new attribute, the add-on's config vars become available in the same `terraform apply` as the add-on is created.

### DIFF
--- a/heroku/resource_heroku_addon.go
+++ b/heroku/resource_heroku_addon.go
@@ -188,10 +188,21 @@ func resourceHerokuAddonRead(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	configVarValues, err := retrieveSpecificConfigVars(client, addon.App.Name, addon.ConfigVars)
+	var configVarValues map[string]string
+	err = resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
+		configVarValues, err = retrieveSpecificConfigVars(client, addon.App.Name, addon.ConfigVars)
+		if err != nil {
+			return resource.NonRetryableError(err)
+		}
+		if len(configVarValues) != len(addon.ConfigVars) {
+			return resource.RetryableError(fmt.Errorf("Got %d add-on config vars from the app, but expected %d", len(configVarValues), len(addon.ConfigVars)))
+		}
+		return nil
+	})
 	if err != nil {
 		return err
 	}
+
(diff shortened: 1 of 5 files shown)


heroku/resource_heroku_addon.go
```diff
@@ -269,3 +290,26 @@ func AddOnStateRefreshFunc(client *heroku.Service, appID, addOnID string) resour
...
+func retrieveSpecificConfigVars(client *heroku.Service, appID string, varNames []string) (map[string]string, error) {
+	vars, err := client.ConfigVarInfoForApp(context.TODO(), appID)
+	if err != nil {
+		return nil, err
+	}
+
+	nonNullVars := map[string]string{}
+	for k, v := range vars {
+		isSpecified := false
+		for _, specifiedVar := range varNames {
```
davidji99, line 303: You can leverage https://github.com/heroku/terraform-provider-heroku/blob/master/heroku/helper.go#L90-L97 to check if `specifiedVar` is in `varNames`.

heroku/resource_heroku_addon.go
```diff
@@ -176,6 +188,15 @@ func resourceHerokuAddonRead(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
+	configVarValues, err := retrieveSpecificConfigVars(client, addon.App.Name, addon.ConfigVars)
```
davidji99, line 191: Is the behavior of the addon config var values being present on the app consistent after the addon is initially provisioned? Do we need some sort of loop/timeout to account for edge cases where the addon config var values are not present just yet?

heroku/resource_heroku_addon.go
mars: Nice catch! Refactored to use that helper.

heroku/resource_heroku_addon.go
mars: I didn't observe that, but agree that a race condition seems likely 😅 so added a retrier on create to await all the add-on vars being set.

davidji99 (approved):

PR #330, opened 2022-01-28
Accelerate Acceptance Tests

mars:
Run all space-related tests together, **using a single space**, factoring-out unnecessary space usage.

What ties it all together is the extended sequence of test steps in `TestAccHerokuSpace`:
https://github.com/heroku/terraform-provider-heroku/compare/test-with-single-space?expand=1#diff-01891d01718a55b6a25566de5ec1e21120ab5e40927361fe30bff8ddcf13c5c3R35-R42

These call to the pre-existing space-dependent tests steps that used to be entire tests that each created & destroyed its own space.

Along the way, I optimized numerous other tests:
* removed unnecessary space requirement from several Heroku App tests
* removed the Shield Space test, because really that just tests setting the `shield` attribute on the space.


mars:
Acceptance test suite run time reduced to 35-minutes!

davidji99 (approved):

PR #331, opened 2022-01-31
Fix hashicorp image link in docs

SriMethan:


SriMethan:
@heroku @davidji99 @mars

davidji99 (approved):

PR #332, opened 2022-02-03
Apps by ID

### LINKED ISSUE #93: heroku_addon_attachment uses app_id instead of app

### LINKED ISSUE #124: Renaming an App breaks many things

### LINKED ISSUE #311: Addon attachments uses invalid addon identifier, forces replacement
### Terraform + Heroku Provider Version
Run `terraform -v` to show core and any provider versions. A sample output could be:

```
Terraform v0.15.3
on darwin_amd64
+ provider registry.terraform.io/heroku/heroku v4.4.1
```

### Affected Resource(s)
- heroku_addon_attachment

```hcl-terraform
resource "heroku_addon" "cache" {
  app  = heroku_app.api.name
  plan = "heroku-redis:hobby-dev"
}

resource "heroku_addon_attachment" "cache" {
  app_id   = heroku_app.api.id
  addon_id = heroku_addon.cache.id
  name     = "REDIS_CACHE"
}
```

### Output
```
  # heroku_addon_attachment.cache must be replaced
-/+ resource "heroku_addon_attachment" "cache" {
      ~ addon_id  = "756ae678-6302-49ef-9be7-fdf827b70fc3" -> "redis-transparent-46729" # forces replacement
      ~ id        = "662702bd-c05e-41b6-a8ff-09aa2edde56d" -> (known after apply)
        name      = "REDIS_CACHE"
      + namespace = (known after apply)
        # (1 unchanged attribute hidden)
    }
```

### Expected Behavior
I expect that the `addon_id` is either the name of the addon or the identifier.

### Actual Behavior
When I set it to its UUID (like mentioned in the docs), it tries to replace it with the name. Then when I run apply, it stores the UUID in the state. So Terraform tries to replace the attachment everytime it runs.

### Steps to Reproduce
1. `terraform apply`

mars:
Fixes #311 
Fixes #124 
Fixes #93

App `name` & `id` are interchangeable in the Heroku API, as a human-friendly convenience. This Terraform Provider was originally implemented using app names as the identifiers. See [these comments](https://github.com/heroku/terraform-provider-heroku/blob/v4.9.0/heroku/resource_heroku_app.go#L192-L200) for some history. This has become mixed-up over the years, as some resources really do require app ID.

The interchangeability of these identifiers is not a well documented or explained situation, especially in the context of Terraform. Really, **this provider should have never allowed using app names as identifiers… hindsight 20/20.**

This PR changes:
* `heroku_app` resource to store state `id` as app ID instead of name
* all other resources that reference apps converted to `app_id` attribute (from `app`) and validate that the value is UUID
* docs now explain that app names should not be used for references
* examples were updated to explicitly show app ID references.
 
🚨 **This is a breaking change for existing configurations.** The `heroku_app` resource's state `id` changes from name to UUID. Therefore, we'll need to add a state migration as well, but that still leaves configurations the explicitly set `heroku_app.example.name` as ID references in a broken state.

Therefore, this will be a major-version release which:
* must warn about the breaking nature of upgrading an existing config
* will need clear docs to upgrade configurations and migrate state.


davidji99:
+1 to this long overdue change.

I'd also recommend we do the following:
1. rename `app` attributes to `app_id` to be explicit.
2. add UUID validator the `app_id` so that app name cannot be used.

mars:
State migrations are what remains to be done. I've got a strategy planned for each resource that's been converted to `app_id`:

- check if the pre-existing `app` attribute value is UUID
  - if is UUID, set it in `app_id` 
  - if not UUID, lookup the App via Heroku API, and set its returned ID in `app_id`.

Hoping to get an answer to my [Hashicorp forum question](https://discuss.hashicorp.com/t/migratestate-vs-stateupgrader/35237) so that I implement using the correct SDK version.

ryanbrainard:
I agree this is a good, long overdue change, but from a UX perspective, I was just noticing that both the CLI and Dashboard do not expose the ID (unless using `--json`). I realize they are well outside of this project, but we should consider updating both of those before making this change to ease the transition. Users could also use the `heroku_app` *data source* to achieve something similar right in Terraform, so that might be another approach to document.

mars:
@ryanbrainard I don't think the UX needs know the app ID values, even for migration through this breaking change:

* the state upgraders/migrations for `app`→`app_id` will automatically swap out app names for IDs (work-in-progress, now)
* within Terraform HCL config, it's a matter of swapping `heroku_app.example.name` for `heroku_app.example.id`
* the resource importers that need app-context still take app name (app name-based imports are friendlier, so left them as-is).

None of these require the consumer to lookup app UUIDs, only manually replace `name` with `id` references.

Am I missing somewhere that app IDs would need to be directly input?

mars:
Implemented the automated state migrations. Tested along the way with a local TF config.

Documented the procedure in [UPGRADING](https://github.com/heroku/terraform-provider-heroku/pull/332/files#diff-f411a5009131db5a6c7242b49ca78488f2d274ef9e5e5f4de9262a42108a5386). This will need to be called out clearly in the major release notes. I'd like to add this as a section to the actual published docs; will look into that too.

Ready for review 🙏

docs/data-sources/addon.md
```diff
@@ -22,7 +22,7 @@ output "heroku_addon_data_basic" {
     "Addon from another app",
     "id: ${data.heroku_addon.from_another_app.id}",
     "name: ${data.heroku_addon.from_another_app.name}",
-    "app: ${data.heroku_addon.from_another_app.app}",
+    "app_id: ${data.heroku_addon.from_another_app.app_id}",
```
davidji99, line 25: Might as well update this to newer HCL format.

docs/resources/app.md
```diff
@@ -10,6 +10,8 @@ description: |-
 
 Provides a Heroku App resource. This can be used to create and manage applications on Heroku.
 
+**Always reference apps by ID (UUID) in Terraform configuration**, even though the Heroku API allows Heroku app name & ID to be interchanged as identifiers in API requests. Referencing apps by name will result in configuration churn (recreation of resources), because this provider always sets & expects apps to be identified by ID.
```
davidji99, line 13: I'd a `-> **IMPORTANT!**` to this section to make this call-out even more evident once the page renders on registry.terraform.io.

docs/resources/app.md
```diff
@@ -91,8 +93,8 @@ config vars to `heroku_app_config_association` resource.
 
 The following attributes are exported:
 
-* `id` - The ID of the app. This is also the name of the app.
-* `name` - The name of the application.
+* `id` - The ID (UUID) of the app.
+* `name` - The name of the app.
```
davidji99, line 97: I can't comment on the specific line but we can drop `uuid` from `heroku_app` now that `heroku_app.id` represents the UUID.

davidji99:
> I'd like to add this as a section to the actual published docs; will look into that too.

@mars You can follow this structure on the Newrelic provider: https://github.com/newrelic/terraform-provider-newrelic/tree/main/website/docs/guides

ryanbrainard:
@mars 

> Am I missing somewhere that app IDs would need to be directly input?

It sounds like you've covered all the bases there and should be pretty seamless for users 👍

mars:
I updated docs based on @davidji99's suggestions, moving the **Upgrading** doc into a new **Guides** section.

I've also staged a new version of the Heroku Dev Center article **Using Terraform with Heroku** with upgrading info and revised example configurations. [Preview the article](https://devcenter.heroku.com/articles/using-terraform-with-heroku-staging?preview=1) which will be published after releasing this new major version.

mars:
@davidji99 migration tests are now in-place for `heroku_app`, `heroku_app_feature` (it has a unique composite ID), and the shared helpers `migrateAppToAppID` & `upgradeAppToAppID`.

I think that covers all the feedback 🙌

mars:
Released [v5.0.0-beta.1](https://github.com/heroku/terraform-provider-heroku/releases/tag/v5.0.0-beta.1). It's now [published to the registry](https://registry.terraform.io/providers/heroku/heroku/5.0.0-beta.1) and available for testing.

davidji99 (approved):

mars:
Plan is to merge and release this next week. Please give the `v5.0.0-beta.1` release a test if you can 🙏 and post feedback here.

PR #338, opened 2022-06-16
When a heroku_build fails, query & output the build error

mars:
When a `heroku_build` fails, the existing error message contains a short-lived URL to download the build log. This is a frustrating experience, to have to go fetch the build log separately each time.

**The changeset automatically fetches the URL and sets the complete build log into the error message, so that it appears inline with all other Terraform output.**

Example new `heroku_build` error output, from the the acceptance test case:

```
        Error: Build failed (app: 58203794-bbca-499d-af42-a6c1870d1909), complete build log follows:
        
        -----> Building on the Heroku-20 stack
        -----> Determining which buildpack to use for this app
        -----> Ruby app detected
        -----> Installing bundler 1.17.3
        -----> Removing BUNDLED WITH version in the Gemfile.lock
        -----> Compiling Ruby
        
         !
         !     There was an error parsing your Gemfile, we cannot continue
         !     
         !     #  from /tmp/build_25aef0c6/Gemfile:3
         !     #  -------------------------------------------
         !     #
         !     >  <<<<<<< Intentional syntax error to break the build
         !     #
         !     #  -------------------------------------------
         !
         !     Push rejected, failed to compile Ruby app.
        
         !     Push failed
        
        (End of build log)
```

Also includes a Terraform plugin SDK upgrade, because it was throwing `error: unsupported state format version: expected ["0.1" "0.2"], got "1.0"`.


mars:
I'll be replacing the Heroku service account & team GitHub Action secrets this week, so that CI can run successfully again.

davidji99 (approved):

PR #340, opened 2022-08-08
Remove erroneous "optional" because this is readonly attribute

mars:
While working with the provider, we realized that the read-only, computed `heroku_addon.config_var_values` is marked `optional: true`, which means Terraform will accept configuration for it.

This changeset removes the `config_var_values` attribute's optional write-ability, which will cause configurations that may have set this (with no effect) to throw an error for the unsettable attribute.


wchrisjohnson (approved):
lgtm

PR #345, opened 2022-10-11
Document the format required by "headers"

mars:
This PR simply documents what is expected for the provider's `headers` attribute.

In helping solve an internal usage question, we realized that this is only documented by [a test](https://github.com/heroku/terraform-provider-heroku/blob/v5.1.4/heroku/provider_test.go#L67).


wchrisjohnson (approved):
lgtm

davidji99 (approved):

PR #346, opened 2022-10-19
Upgrade module to v5

sarahoh2468:
The tagging and module version are not consistent (currently on v4 and most recent release is 5.1.5). This PR upgrades the module to v5

More context here: https://salesforce-internal.slack.com/archives/CCHLHGYN4/p1666192706879899

(For the public, this change is to resolve an error with importing this provider as a module for use within another provider.)

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -18,7 +18,7 @@ test: fmtcheck
 		xargs -t -n4 go test -v $(TESTARGS) -timeout=30s -parallel=4
 
 testacc: fmtcheck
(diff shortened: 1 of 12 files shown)


mars:
@sarahoh2468 why not only update the paths to `v5` instead of a whole new major revision?

mars (approved):
Thank you @sarahoh2468, this looks good ✅

Once acceptance tests pass (approx 30-minutes wait time), I will merge & release as `v5.1.6`.

PR #362, opened 2023-02-21
Omit sensitive_config_vars values from TF_LOG=debug output

mars:
Discovered this disclosure vulnerability when debugging a Terraform configuration with `TF_LOG=debug`.

This changes the behavior to output just the `heroku_app` `sensitive_config_vars` names and not their values.


wchrisjohnson (approved):
Love this change! Protects the sensitive info but still gives feedback on what has changed.

I wish every provider did this.

PR #367, opened 2023-03-14
Fix build flag to set provider version for user-agent string

mars:
This build flag has been incorrect for years, and as a result the provider sends requests as user-agent `terraform-provider-heroku/dev` for all usage that is not [an acceptance test](https://github.com/heroku/terraform-provider-heroku/blob/master/GNUmakefile#L21), instead of with the version, such as `terraform-provider-heroku/5.1.1`.

This corrects the build flag to set the version correctly as [documented in the comments](https://github.com/heroku/terraform-provider-heroku/blob/master/version/version.go#L7).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   goreleaser:
-    runs-on: ubuntu-latest
+    runs-on: pub-hk-ubuntu-22.04-small
     steps:
       - 
         name: Checkout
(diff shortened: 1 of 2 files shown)


davidji99 (approved):

mars:
Finally confirmed that this is working with [v5.1.13.pre.3](https://github.com/heroku/terraform-provider-heroku/releases/tag/v5.1.13-pre.3) by observing internal Heroku API usage analytics, when the versioned user-agent finally appeared:
```
heroku/v5 (darwin; amd64) terraform-provider-heroku/5.1.13-pre.3
```

PR #368, opened 2023-04-10
Skip heroku_review_app_config.deploy_target.id region lookup for space

ryanbrainard:
The current code assumes that the `heroku_review_app_config.deploy_target` is of type `region` and tries to resolve it to a name, but in the case of a deploy target that is a `space`, it should remain as the `id`, as space names would violate the schema and cause breaking changes. This skips the name look up if the type is space. Without this code change, Terraform updates to `heroku_review_app_config` are made in Heroku but then fail to update the Terraform state.

Note, I did not add tests to this because the current tests depend on a real pipeline in CI, which I'm not sure has spaces enabled and I don't have access to check or configure it. I have been able to manually test this however with my own pipeline.

### DIFF
--- a/heroku/resource_heroku_review_app_config.go
+++ b/heroku/resource_heroku_review_app_config.go
@@ -12,6 +12,11 @@ import (
 	heroku "github.com/heroku/heroku-go/v5"
 )
 
+const (
+	DeployTargetTypeSpace = "space"
+	DeployTargeTypeRegion = "region"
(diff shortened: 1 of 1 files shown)


mars (approved):
This looks great @ryanbrainard 👍 Thank you for filling this gap in functionality, so that terraformed Review App Config works for Review Apps in Private Spaces.

PR #369, opened 2023-04-11
Spell target correctly

ryanbrainard:
I accidentally spelled "target" wrong in https://github.com/heroku/terraform-provider-heroku/pull/368. This fixes that.


mars (approved):
👍 We'll go head and merge. It will be included in the next release.

PR #373, opened 2023-06-15
Fix Private Space create without DataCIDR

mars:


voidlock (approved):
LGTM, thank you for fixing

PR #389, opened 2024-09-06
Add customDiff for sensitive config vars

AlfonsoCampodonico:


mars:
This seems like a good approach, but the Custom Diff must account for all attributes that contribute to the resource's configurable state, not just `sensitive_config_vars`.

heroku-johnny:
Since there has been no response or activity on this PR for over a year I'm going to close this for now. If this is needed we can address this at a later date

PR #436, opened 2026-06-23
Add CNB support for Cedar-generation apps

heroku-johnny:
## Description

Cedar-generation apps can now opt into Cloud Native Buildpacks by setting `stack = "cnb"`. Previously, CNB was exclusive to Fir-generation apps. This change extends CNB support to Cedar while keeping traditional buildpacks working for existing Cedar apps that don't set `stack = "cnb"`.

Related: W-23144558

## Changes

- Add `IsCNBApp(generation, stack string) bool` helper — canonical check for whether an app instance is using CNB
- Update feature matrix: Cedar `cloud_native_buildpacks` is now `true`
- Guard `BuildpackInstallationList` call in app resource against Cedar CNB apps (the API returns a 422 for CNB apps; we skip the call entirely)
- Make buildpack validation in build resource stack-aware — Cedar apps with `stack = "cnb"` now reject traditional buildpacks at plan and apply time
- Remove unused `isCNBError` helper

## Usage

```hcl
  name   = "my-cedar-cnb-app"
  region = "us"

  # Opts this Cedar app into Cloud Native Buildpacks.
  # Configure buildpacks via project.toml, not the buildpacks attribute.
  stack = "cnb"
}
```

Existing Cedar apps using traditional buildpacks are unaffected.

### DIFF
--- a/docs/resources/app.md
+++ b/docs/resources/app.md
@@ -36,6 +36,22 @@ resource "heroku_app" "cedar_app" {
 }
 ```
 
+### Cedar Generation Using Cloud Native Buildpacks
+```hcl-terraform
+resource "heroku_app" "cedar_cnb_app" {
+  name   = "my-cedar-cnb-app"
+  region = "us"
+
+  # Setting stack = "cnb" opts this Cedar app into Cloud Native Buildpacks.
+  # Do not specify buildpacks — configure them via project.toml instead.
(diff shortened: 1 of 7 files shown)


mars (commented):
Nice one! A few suggestions.

mars (approved):
👏👏👏

